PR #20, opened 2026-02-01
feat: implement unified version management

HoneyBBQ:
## Summary
This PR introduces a unified version management standard by centralizing version information into a shared `internal/version` package, moving away from hardcoded variables in individual `main` packages.

Key Changes:
- **New `internal/version` package**: Supports build-time injection of version, commit hash, and build time via `ldflags`.
- **Modern Fallback Mechanism**: If the commit hash is not injected at build time, it automatically attempts to retrieve the Git revision from the binary's build metadata using `runtime/debug` (Go 1.18+).
- **Component Refactoring**:
  - `cmd/mcp`: Removed legacy hardcoded version logic and updated Dockerfile to inject metadata into the new package path.
  - `cmd/agent`: Added version information to the startup logs.
  - `cmd/cli`: Added a new `--version` flag for users to check the CLI version.

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -3,13 +3,14 @@ package main
 import (
 	"context"
 	"fmt"
-	"log"
+	"log/slog"
 	"os"
 	"strings"
 	"time"
 
 	"github.com/memohai/memoh/internal/chat"
 	"github.com/memohai/memoh/internal/config"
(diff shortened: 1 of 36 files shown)


HoneyBBQ:
> The currently-use cli is at /packages/cli build by TypeScript, may should add --version command to it

Done.

PR #75, opened 2026-02-20
chore(husky): large file pre-commit check support

Fodesu:
max file size `MAX_SIZE` now is 1M, does we have better magic number?


Fodesu:
rebase 把这个 pr 搞的好乱...

PR #344, opened 2026-04-08
fix: preserve attachment base64 for telegram fallback

kuishou68:
$## Summary\n- preserve attachment base64 data when replaying streamed attachment deltas\n- let Telegram file sending fall back to embedded base64 when platform keys and content hashes are unavailable\n- add regression coverage for attachment delta parsing and Telegram video fallback\n- fix the attachment delta regression test to pass JSON into parseAttachmentDelta\n\nCloses #311


kuishou68:
> Fixed in #349.

……

PR #377, opened 2026-04-14
feat(skills): add effective skill resolution and actions

### LINKED ISSUE #366: [Feature]: agent-managed skill directories and sync with Web console
### Feature Description

Please consider adding first-class support for agent-managed skills so that skills installed/updated inside the workspace can be reliably discovered and reflected in the Web console.

### Motivation

Memoh is increasingly agent-native: bots can manage files, tools, and runtime state inside their own workspace. Skills should follow the same model.

Right now, skill installers from different ecosystems may write to different directories depending on version, client, or runtime environment:
  - Memoh Web console could manage skills.md under `/data/.skills`
  - vercel-lab/skills (`npx skills ...`) will write to `/data/.agent/skills`
  - lobehub/skills (`npx @lobehub/market-cli skills`) will write to `/root/.openclaw/skills`

This can cause drift between:
  - what the agent can actually use
  - what the Web console can actually see/manage

### Proposed Solution

A solution in this area could include one or more of:
  - configurable skill roots via environment variables
  - symlink/canonical-path support
  - multi-directory discovery with precedence
  - filesystem-to-Web sync or refresh mechanism
  - UI metadata showing source/effective status of discovered skills

### Alternatives Considered

- for vercel-lab/skills (`npx skills ...`) there isn't any methods to customize skills directory
- for lobehub/skills (`npx @lobehub/market-cli skills`) could add `--dir` every time to let it save to Memoh default skills directory

### Desired outcome

Memoh should have a consistent “effective skills” view, where:
  - agent-installed skills are discoverable
  - Web console stays aligned with runtime state
  - backward compatibility across older/newer skill directory layouts is preserved (e.g. for https://github.com/memohai/Memoh/commit/bb14bcb3bc50660b1c0602983ca052cf9ad24cfe )

### Contributing

- [x] I would like to submit a pull request to implement this feature

ChrAlpha:
Closes #366

## Summary

This PR introduces a skill discovery and resolution model that keeps the bot container filesystem as the source of truth, while exposing a single effective skill view to both runtime and Web.

Instead of treating every external skill directory as an equal source of truth, this implementation distinguishes between:

- managed skills: Memoh-owned skills under /data/skills
- discovered skills: compatible external skills found in a limited set of known directories (`/data/.agents/skills`, `/root/.agents/skills`)
- effective skills: the resolved set that runtime actually loads
- derived metadata: rebuildable state stored under /data/.memoh/skills/index.json

image

### Backend

- Adds a dedicated internal/skills package for:
    - skill discovery
    - frontmatter parsing
    - effective-state resolution
    - derived index read/write
    - skill actions: adopt, disable, enable
- Keeps /data/skills as the canonical managed directory
- Adds compatibility discovery for a small built-in set of known locations, including:
    - legacy Memoh path
    - selected agent/client-specific paths inside the container
- Introduces explicit skill source metadata:
    - source_path
    - source_root
    - source_kind
    - managed
    - state
    - shadowed_by
- Makes runtime load only the effective skill set
- Adds a skill action API endpoint for adopt/disable/enable
- Updates supermarket skill installation to use the managed directory

### Workspace / Runtime

- Normalizes container env for new workspaces:
    - HOME=/data
    - XDG_CONFIG_HOME=/data/.config
    - XDG_DATA_HOME=/data/.local/share
    - XDG_CACHE_HOME=/data/.cache
- Updates prompt copy to reflect the managed skill directory and compatibility discovery
  behavior

### Web

- Shows skill provenance and effective state in the Skills page
- Distinguishes:
    - Managed vs Discovered
    - Effective / Shadowed / Disabled
- Supports:
    - Adopt into Memoh-managed skills
    - Disable
    - Enable
- Restricts delete to managed skills only
- Uses the real source_path when opening skill files from the session info panel
- Includes a small follow-up UI polish pass for:
    - header alignment
    - visibility icon semantics
    - more compact metadata layout on narrow widths

> A skill source is **shadowed** when another source with the same skill name has higher priority and is currently the effective one. The shadowed source is still discovered and visible in the UI, but it is not loaded into runtime unless the higher-priority source is removed or disabled.
>
> **Adopt** means copying a discovered external skill into Memoh’s managed skill directory, so Memoh becomes the owner of that skill content. After adoption, the managed copy becomes the effective source, while the original external source remains visible as a discovered source and is typically shown as shadowed.

---

P.S. I'm new to Go and these may not be the best practice, so I'm open to discuss whether we want to move forward with this implementation.


sheepbox8646:
也许可以在 `use_skill` 的工具返回中增加 `path`

比如 `use_skill(skill="pdf", reason="xxx")`

返回 `path="/data/.agents/skills/pdf"`

sheepbox8646 (approved):

PR #382, opened 2026-04-16
fix: exclude expired token adjustments from usage stats

zzj3720:
## What
- exclude negative token usage rows from daily usage aggregation
- exclude negative token usage rows from by-model usage aggregation
- keep expired token adjustment rows out of the frontend usage response

## Validation
- `mise run sqlc-generate`
- `go test ./internal/handlers ./internal/command ./internal/db/sqlc`
- manual PostgreSQL verification with one positive usage row and one negative expired-adjustment row; the aggregated result kept only the positive row

### DIFF
--- a/db/queries/token_usage.sql
+++ b/db/queries/token_usage.sql
@@ -1,40 +1,80 @@
 -- name: GetTokenUsageByDayAndType :many
+WITH usage_rows AS (
+  SELECT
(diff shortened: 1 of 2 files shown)


zzj3720:
Closing this PR because I targeted the wrong repository for the Slack task.

PR #395, opened 2026-04-22
fix(web): show attachment messages immediately

zzj3720:
## Summary
- move attachment file reading out of the pre-send path so the optimistic user turn renders immediately
- use blob preview URLs for optimistic attachment rendering and resolve the real base64 payload right before transport
- revoke temporary blob URLs after refresh/send completion and allow blob URLs in the media gallery

## Testing
- corepack pnpm exec eslint apps/web/src/composables/api/useChat.message-api.ts apps/web/src/composables/api/useChat.types.ts apps/web/src/pages/home/components/chat-area.vue apps/web/src/pages/home/composables/useMediaGallery.ts apps/web/src/store/chat-list.ts
- corepack pnpm --filter @memohai/web exec vue-tsc --noEmit -p tsconfig.json
- corepack pnpm --filter @memohai/web build
- manual validation on a minimachine Safari session against a mock Memoh server; injected a 3s FileReader delay and confirmed the message text plus attachment appeared in 0.52s after clicking send

### DIFF
--- a/apps/web/src/composables/api/useChat.message-api.ts
+++ b/apps/web/src/composables/api/useChat.message-api.ts
@@ -54,6 +54,36 @@ export interface SendMessageOverrides {
   reasoningEffort?: string
 }
 
+function readFileAsDataUrl(file: File): Promise<string> {
+  return new Promise((resolve, reject) => {
+    const reader = new FileReader()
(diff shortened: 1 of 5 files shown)


zzj3720:
Closing this PR because the requested fix belongs in Cueboard rather than Memoh. I am moving the work to Cueboard.

Issue #386, opened 2026-04-17
[Bug]:  Failed to install memoh using one click script


D0ot:
### Bug Description

As title, see the reproduce steps for detail.

### Steps to Reproduce

These are my install steps.
```
localserver0:~/apps$ curl -fsSL https://memoh.sh | sudo sh
Memoh One-Click Install
✓ Docker and Docker Compose detected
✓ Latest release: v0.7.0
✓ Docker image version: 0.7.0
Configure Memoh (press Enter to use defaults):

  Workspace (install and clone here) [~/memoh]: 
  Data directory (bind mount for server container data) [/root/memoh/data]: 
  Admin username [admin]: admin
  Admin password [admin123]: admin@memoh0
  JWT secret [auto-generated]: 
  Postgres password [memoh123]: admin@memoh0
  Enable sparse memory service? [y/N]: y

  Browser core selection:
    1) Chromium only (default, smaller image)
    2) Firefox only
    3) Both Chromium and Firefox
  Browser core [1]: 1

Cloning Memoh into /root/memoh...
Cloning into 'Memoh'...
remote: Enumerating objects: 2071, done.
remote: Counting objects: 100% (2071/2071), done.
remote: Compressing objects: 100% (1730/1730), done.
remote: Total 2071 (delta 334), reused 1330 (delta 253), pack-reused 0 (from 0)
Receiving objects: 100% (2071/2071), 8.65 MiB | 4.11 MiB/s, done.
Resolving deltas: 100% (334/334), done.
Note: switching to 'd3a820b2dc4b4d91cf752d4312f62f2dc30fa9a3'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

✓ Docker images pinned to 0.7.0
✓ Sparse memory service enabled
✓ Browser: chromium (image tag: 0.7.0-chromium)

Pulling Docker images...
[+] pull 6/6
 ✔ Image qdrant/qdrant:latest           Pulled                                                                                                                                                                                                             4.7s
 ✔ Image postgres:18-alpine             Pulled                                                                                                                                                                                                             5.0s
[...]

PR #401, opened 2026-04-23
fix(install): handle dirty one-click upgrades

ChrAlpha:
Follow https://github.com/memohai/Memoh/issues/386#issuecomment-4296173521

## Summary

The install script now detects existing Memoh config, env files, Docker volumes, and containers before deciding whether to perform a fresh install, upgrade, or reinstall. Silent reruns with an existing `config.toml` now default to upgrade and preserve database/admin credentials instead of regenerating mismatched secrets. Dirty Docker state without reusable config now fails fast unless the user explicitly chooses reinstall.

## Changes

- Add `MEMOH_INSTALL_MODE` / `--install-mode` with `auto`, `fresh`, `upgrade`, and `reinstall`
- Reuse existing `config.toml` and `.env` values during upgrades
- Refuse unsafe fresh installs when existing Memoh Docker state is detected
- Add explicit reinstall cleanup for Memoh containers, volumes, and network
- Overwrite `.env` instead of appending duplicate values on reruns
- Print recent postgres/migrate logs when compose startup fails
- Document upgrade/reinstall behavior in Docker install docs

### DIFF
--- a/docs/docs/installation/docker.md
+++ b/docs/docs/installation/docker.md
@@ -65,7 +65,7 @@ The script will:
 6. Fetch the latest release tag from GitHub and clone the repository
 7. Generate `config.toml` from the Docker template with your settings when needed
 8. Pin Docker image versions to the release
-9. Build the browser image with selected cores and start all services
+9. Select and pull the prebuilt browser image for the chosen cores and start all services
 10. Print recent `postgres` and `migrate` logs automatically if startup fails
 
 **Silent install** (use all defaults, no prompts):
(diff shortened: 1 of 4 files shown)


firavoyage:
@claude @copilot 

ive seen `docker compose down -v` in the markdown.

how could it keep the data even with this command? does it move the data somewhere as backup and copy back later?

for all cases, this command should not run, wo checking "there is no data!" or backup.

though, ive not found actual reference in code. seems it just reuses the prev config.

also, it mainly targets the problem of the one liner running twice. but actually it could be installed anywhere (you can even choose where to install inside the one liner, wo build from source. idk if it saves or exports somewhere for the future.). since you wont scan the whole disk, you have no idea where it is.

ive read the code. whenever it does not find one, it just attempts to clear all your data.

just confusing.

sheepbox8646 (approved):

PR #487, opened 2026-05-16
remove an unusable button.

firavoyage:
remove an unusable button.

`href = #` does nothing but surprises the user.

rather remove it (or place a todo) than having a broken link there.


alexma233:
Maybe you can remove i18n locales of `auth.forgotPassword` if no page uses it.

sheepbox8646 (approved):

PR #527, opened 2026-05-29
perf(web): virtualize model picker and searchable select dropdowns

qqqqqf-q:
## Summary
- Virtualize the model picker (`model-options.vue`) and the shared `searchable-select-popover` (used by ~10 selects: timezone, TTS model, provider selects, access control, chat language…) with `@tanstack/vue-virtual`. Providers like OpenRouter expose hundreds of models and timezone lists run to hundreds of entries; previously every option mounted at once on open, freezing the main thread and dropping the popover open animation. Now only visible rows mount, with row heights measured at runtime.
- Fix a Tailwind v4 width util (`w-[--var]` no longer auto-wraps in `var()`).
- Scope the narrower popover width (0.55× trigger) to the timezone picker only, via a `widthRatio` prop defaulting to full width — the other ~9 consumers are unaffected.

## Dependency note
Promotes `@tanstack/vue-virtual` to a direct dependency. It is **already in the tree transitively via reka-ui** (`reka-ui@2.7.0` → `^3.12.0`); pnpm dedupes to a single version `3.13.17`, so this adds **zero** new package to the bundle — it only makes the existing dependency explicit.

## Test plan
- [ ] Model picker (bot settings / chat pane) with a provider exposing hundreds of models: opens instantly, smooth scroll, search + selection work.
- [ ] Timezone picker (new bot / bot Global card / profile): narrow (0.55) width, hundreds of zones scroll smoothly, search + select work.
- [ ] Other searchable selects (TTS model, provider selects, access control, chat language): full trigger width restored; open/scroll/select correct.
- [ ] Scrollbar tracks correctly (no drift) in long lists.

## Known limitations (follow-up)
- Off-screen options leave the DOM, so Tab-key / screen-reader users can't reach options outside the rendered window, and the selected item isn't auto-scrolled into view on open. Deferred.
- `settings-interaction-card.vue` still uses the un-wrapped `w-[--reka-popover-trigger-width]` (same Tailwind v4 issue) — out of scope, separate fix.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -26,6 +26,7 @@
     "@shikijs/transformers": "^4.0.1",
     "@tailwindcss/vite": "^4.2.2",
     "@tanstack/vue-table": "^8.21.3",
+    "@tanstack/vue-virtual": "^3.12.0",
     "@vee-validate/zod": "^4.15.1",
     "@vueuse/core": "^14.1.0",
     "@vueuse/integrations": "^14.2.1",
--- a/apps/web/src/components/searchable-select-popover/index.vue
+++ b/apps/web/src/components/searchable-select-popover/index.vue
@@ -28,7 +28,8 @@
       </slot>
     </PopoverTrigger>
     <PopoverContent
-      class="w-[--reka-popover-trigger-width] p-0"
+      class="p-0"
+      :style="{ width: 'calc(var(--reka-popover-trigger-width) * 0.55)' }"
       align="start"
     >
       <div class="flex items-center border-b px-3">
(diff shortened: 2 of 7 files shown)


qqqqqf-q:
@codex review

Added keyboard navigation for the virtualized dropdowns to address the two P2 findings. Both the model picker and the shared searchable-select popover now support a combobox keyboard pattern (Arrow/Home/End to move an active option, Enter to select), with the active row scrolled into view via the virtualizer and exposed through aria-activedescendant — so keyboard and screen-reader users can reach off-screen options again. Implemented as a shared useListboxKeyboard composable.

qqqqqf-q:
@codex review

chen-ran (approved):

PR #553, opened 2026-06-02
feat(command): command UX revamp — Telegram interactivity + i18n + no-button platform parity

qqqqqf-q:
## Summary

- **Telegram command UX**: inline buttons, pagination, model picker, drill-down sub-command navigation, edit-in-place, collapse-on-close, action menus for bare groups, empty-state buttons. Also auto-registers the bot's command menu with Telegram via `setMyCommands` on startup.
- **Command UI localization**: new `command_ui_language` per-bot setting (independent of chat language). English + Simplified Chinese catalogs. Many hardcoded English strings in command outputs (model errors, schedule errors, tool-approval errors, /mcp back button, /help separators, etc.) now flow through i18n. A persistent schema change — adds the `bots.command_ui_language` column (SQLite 0011, Postgres 0084).
- **No-button platform parity**: the 11 channel adapters without `Capabilities.Buttons` (Discord, Slack, Feishu, DingTalk, QQ, Matrix, Misskey, WeCom, Weixin, WeChat OA, Local-Web) now receive auto-derived typeable-command guidance for every Interactive surface. Per-surface UX fixes for cases that previously stranded users (`/model` showed 12-of-564 unrelated models with no provider breakdown; `/usage` rows collapsed into one paragraph; `/fs list` directory tree collapsed).
- **Plain-text channel silent-failure fix at the outbound boundary**: `channel.CoerceFormatForCaps` runs right before `validateMessageCapabilities` and degrades auto-promoted Markdown back to Plain (with inline markup stripped) on channels that can't render Markdown — instead of letting validation reject the message (logged but never replied, the original WeChat silent failure). The defense is now a property of the outbound boundary rather than per-call-site discipline.
- **Shared formatting layer (`humanize.go`)**: new ~330-line module the command outputs use for time-ago, byte sizes, durations, cron strings, status — replaces ad-hoc `time.Format` calls scattered across handlers and gives Telegram lists their consistent compact rendering.

## Why

Of the 12 channel adapters, only Telegram declares `Capabilities.Buttons: true`. The other 11 platforms (~92% of supported surface area) silently dropped the entire `Interactive` payload — leaving users with text bodies that often had no path forward. Half of the 16 Interactive sites stranded text-channel users entirely. The branch's prior work made the Telegram experience much better; this revamp closes the gap so the other 11 platforms are first-class too.

## What changes for users

- **Telegram**: still gets the full interactive picker, drill-down, etc. Text bodies on a handful of surfaces are slightly cleaner — previously they baked typing hints alongside live buttons; now those hints only appear when buttons aren't available. The bot menu is registered with Telegram automatically (no more manual `/setcommands`).
- **Discord / Slack / Feishu / DingTalk / QQ / Matrix / Misskey / WeCom**: now see a structured typeable-command trailer for every Interactive surface (e.g. `Open:\n- /reasoning show\n- /model list\n- /settings update --heartbeat_enabled true` after `/settings`).
- **Weixin / WeChat OA / Local-Web**: commands no longer silently fail. Multi-line content (`/usage`, `/fs list`, `/help <group>`, settings card, etc.) uses bullet formatting that survives plain-text IM soft-wrap collapse.
- **`/model` on no-button channels**: provider summary (`Chat Models (564) — DeepSeek (4) ●, OpenAI (124), OpenRouter (436)`) instead of a flat 12-of-564 list.
- **`/reasoning show` on no-button channels**: now includes the explainer "Higher effort means more careful thinking, but slower replies. Choose a level:" in the body.
- **`/settings language` (and `/language`) on no-button channels**: shows a `Current: <label>` line under the picker title.

## Architecture (new pieces)

- **`internal/command/fallback.go`** derives a typeable-command trailer from the structured `Interactive` payload (`ItemAction{Resource, Action, Args}`) via a 7-verb i18n dictionary (`cmd.fallback.{switch,pick,toggle,open,details,range,menu}`).
- **`type HintVerb string`** with seven constants — typed string (not raw) so adding a verb is a two-step change: register the constant + add the localized template.
- **Optional override field**: `ListView.HintVerb`. Empty = inferred from structure. Unknown verb gracefully degrades to inference rather than emitting an empty trailer. (Row-level override `ItemAction.Verb` was removed during review — no production caller set it.)
- **`ChoicesView.BodyEnumeratesChoices bool`**: contract-named flag for surfaces whose body Text already enumerates every typeable affordance (e.g. `/help <group>`'s Usage block) — skip trailer derivation.
- **`(*ItemAction).Typeable()`** and **`ParsedCallback.SyntheticCommand()`** both go through a shared `formatSlashCommand(resource, action, args, quote)` helper, so the user-facing display path (quoted args, copy-paste safe) and the internal re-dispatch path (raw args, page suffix) can't drift in their slash-command encoding.
- **`channel.CoerceFormatForCaps`** (new) and **`channel.StripInlineMarkup`** (newly exported): the outbound boundary degrades Markdown→Plain when the target channel can't render Markdown, instead of relying on every caller upstream to explicitly set `Format=Plain`.
- **Renderer hook** at `internal/channel/inbound/result_render.go:appendFallbackTrailer` — fires only when `!Caps.Buttons && Interactive != nil`. Nil-view safety: `renderModelPicker` / `renderChoicesView` take a `fallbackText` parameter and return it when the view is nil rather than crashing on a partial `Interactive`.
- **`internal/command/humanize.go`** (new) — shared formatters used across `/usage`, `/fs`, `/schedule`, `/heartbeat`, `/mcp`, `/memory`, `/search` list/get outputs. Time-ago, durations, byte sizes, status text, cron pretty-print.

## Migrations

- **SQLite**: `0011_command_ui_language.{up,down}.sql` (renumbered from 0009 after reconciling with main's intervening 0008/0010). Main's 0010 rebuilds the `bots` table without `command_ui_language`, so this branch's column-add now sequences AFTER the rebuild rather than being silently dropped.
- **Postgres**: `0084_command_ui_language.{up,down}.sql` — pure `ADD COLUMN IF NOT EXISTS`; no rebuild collision because main's 0085 is `ALTER COLUMN ... SET DEFAULT` only.

## Test plan

- [x] `go test -race ./internal/command/... ./internal/channel/inbound/... ./internal/channel/... ./internal/db/... ./internal/settings/...`
- [x] Coverage gate (`internal/channel/inbound/fallback_coverage_test.go`): every Interactive kind × override flag must produce a non-empty trailer with a `/command` and no `--page 0` artifact; mixed-shape nil-view cases pinned.
- [x] Cross-platform consistency (`internal/channel/inbound/cross_platform_consistency_test.go`): the three capability profiles (plain-text / markdown / richtext) must produce equivalent content after markup strip.
- [x] Typeable round-trip (`TestTypeableRoundTripsThroughParse`): every `Typeable()` output re-parses to the same `(Resource, Action, Args)` — protects copy-paste correctness for args with whitespace.
- [x] Outbound-boundary defense (`TestCoerceFormatForCaps_*`): pins the Markdown→Plain auto-degrade on plain-text-only channels, and asserts Plain stays Plain everywhere + Markdown stays intact on capable channels.
- [x] Telegram button-tap round-trip (`TestTelegramCallbackTapRoundTrip`): joins `renderResult → DecodeCallback → SyntheticCommand → Parse` end-to-end (the chain a user's button tap travels through).
- [x] Full SQLite migrate up → down cycle (1→11→1) clean locally.
- [x] Manual: `/help`, `/settings`, `/model`, `/usage`, `/reasoning`, `/memory`, `/heartbeat`, `/email` on Telegram (full interactive flow) and WeChat (plain-text fallback).

### i18n key renames (callers maintaining custom translations need to update)

- `cmd.settings.btn*` → `cmd.settings.section.*` / `cmd.settings.action.*`
- `cmd.email.btn*` → `cmd.email.section.*`
- `cmd.heartbeat.btnSettings` → `cmd.heartbeat.section.settings`
- `newSession.btn{Confirm,Cancel}` → `newSession.action.{confirm,cancel}`
- `cmd.help.tapAction` → `cmd.help.chooseAction`
- `cmd.reasoning.tapPrompt` → `cmd.reasoning.choosePrompt`

### i18n keys moved (same values, different location)

- `cmd.schedule.fieldCreated` → `cmd.common.fieldCreated`
- `cmd.schedule.fieldUpdated` → `cmd.common.fieldUpdated`
- `cmd.mcp.fieldStatus` and `cmd.schedule.fieldStatus` → `cmd.common.fieldStatus`

### New API surface

- `command.HintVerb` (typed string constant set)
- `command.ListView.HintVerb HintVerb`, `command.ChoicesView.BodyEnumeratesChoices bool` (both optional; zero-value is "infer")
- `command.FallbackTrailer(iv *Interactive, t *i18n.Localizer) string`
- `(*command.ItemAction).Typeable() string`
- `channel.CoerceFormatForCaps(msg Message, caps ChannelCapabilities) Message`
- `channel.StripInlineMarkup(s string) string` (was internal)

### Schema additions

- New column `bots.command_ui_language TEXT NOT NULL DEFAULT 'auto'` (SQLite + Postgres). Existing rows backfilled to `'auto'`. The setting is per-bot, independent of chat-reply language; resolves to server default English at render time when `'auto'`.

image
image


qqqqqf-q:
@codex review

qqqqqf-q:
@codex review

Pushed two follow-up commits:
- `29e053aa` pre-handoff self-review: boundary fixes (callback arg encoding for space-bearing names, zh /model not-found pointer + CJK period, non-owner model picker gating, edit-path terminal-error handling, /new --confirm, callback length gate, i18n parity).
- `e35f8d66` addresses your prior review: /language shorthand arg-plumbing fix, command_ui_language added to both 0001 baselines, and sqlc regenerated (Bot struct was stale).

Please re-review the new state.

qqqqqf-q:
@codex review

`b4c9eebb` addresses your latest P2 (model picker stable identity): the select button now carries the model's stable DB id instead of a positional flat index, so a model-list change between render and tap can't silently switch to a different model. Replaced the `--flat` index flag with `--id` end-to-end; a removed model still yields the "list changed" message.

qqqqqf-q:
@codex review

`0fdbb74d` addresses your latest P2 (provider-qualified model-select hint): when a no-button user drills into a provider's model list, the fallback trailer now suggests the two-arg `/model set <provider> <model_name>` form (provider quoted if it contains spaces) instead of the single-arg `/model set <name>` that errors as ambiguous for names shared across providers.

qqqqqf-q:
@codex review

`7d1276dc` addresses your latest P2 (single-provider no-button case): /model with exactly one provider now drills straight into that provider, so the model-level picker and its `/model set <provider> <model_name>` trailer are attached instead of a LevelProviders `/model list <provider>` hint that led text-only users to a dead end.

qqqqqf-q:
@codex review

`156d86a7` addresses your latest P2 (unnamed models) and proactively sweeps the same empty-label class across the model subsystem: `modelDisplayName` falls back to model_id for blank names (picker candidates, flat list, current-model display), and `resolveProviderName` falls back to the provider id for blank provider names (buttons, summaries, and keeping the provider-qualified `/model set` hint self-consistent).

qqqqqf-q:
@codex review

`f66229e9` addresses your latest P2 (stream-final coercion): the Markdown→Plain downgrade is no longer a side-effect of validateStreamEvent mutating the shared event.Final. Push now coerces a local copy of the final before dispatch, so the adapter receives the downgraded payload and tee fan-out to other channels keeps its own format.

qqqqqf-q:
@codex review

qqqqqf-q:
之后会还有一个 im 使用者权限重构和这个 pr 配合,目前对于新用户而言还是缺少 bind的会有很多东西用不了

qqqqqf-q:
@codex review

qqqqqf-q:
@codex review

sheepbox8646 (approved):

PR #556, opened 2026-06-02
fix(web): use transcription providers in model select

StringKe:
## Summary
- Pass transcription providers into the bot multimedia settings card.
- Use transcription provider metadata for the transcription model select instead of TTS providers.

## Root cause
The transcription model select received `ttsProviders` while rendering transcription models. When a transcription model provider was not present in the TTS provider list, the select fell back to the provider UUID as the group label.

## Validation
- `pnpm --filter @memohai/web build`
- `pnpm exec lint-staged`

Notes: the build still prints the existing lightningcss warnings for `.about-markdown-container :deep(...)` and chunk size warnings; they are outside this change.


sheepbox8646:
@cursoragent review

PR #577, opened 2026-06-06
fix: guard login submission while routing

ChrAlpha:
## Summary

- Prevent duplicate login submissions while an existing login attempt is in progress.
- Keep the login form in a busy/disabled state until post-login navigation finishes.
- Remove the duplicate button click submission path and rely on the form submit handler.
- Add focused tests for login submission idempotency and busy-state behavior.

---

The login page previously showed a spinner during authentication, but the form remained interactive and the submit handler could be triggered again. On successful login, the UI could also become interactive before `router.replace('/')` and route guards completed, which made the several-second transition window race-prone.

This change extracts the login submission flow into a small tested helper, blocks concurrent submissions, disables the username/password inputs and submit button while busy, and awaits navigation before releasing the busy state.

### DIFF
--- a/apps/web/src/pages/login/index.vue
+++ b/apps/web/src/pages/login/index.vue
@@ -1,5 +1,8 @@
 <template>
-  <main class="w-screen h-screen flex *:m-auto bg-background relative p-4">
+  <main
+    class="w-screen h-screen flex *:m-auto bg-background relative p-4"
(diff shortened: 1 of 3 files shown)


ChrAlpha:
@codex review

sheepbox8646 (approved):

PR #580, opened 2026-06-06
fix: close desktop tab with cmd+w

ChrAlpha:
## Summary

- Make macOS `Cmd+W` close the active desktop workspace tab instead of closing the chat window.
- Add a shared typed desktop keyboard command layer between main, preload, and renderer.
- Register workspace-tab closing through the renderer command registry while preserving native close behavior for non-chat windows.


ChrAlpha:
Superseded by #590

PR #590, opened 2026-06-08
feat(keyboard): unify app shortcut commands

ChrAlpha:
## Summary

- Add a shared keyboard command registry and binding table for web and desktop.
- Route desktop `Cmd/Ctrl+W` through the shared close-tab command while preserving native window close fallback.
- Keep browser `Cmd/Ctrl+W` passthrough so the browser tab-close default is not intercepted.
- Move file save shortcut handling (`Cmd/Ctrl+S`) onto the shared keyboard layer with scoped component handlers.

Supersedes #580.

## Notes

This keeps command semantics in `@memohai/web` while letting each surface own delivery:

- Electron menu accelerator for desktop-owned shortcuts.
- Web keydown listener only for browser-safe shortcuts.
- Component-scoped handlers for active/KeepAlive-aware command ownership.

### DIFF
--- a/apps/desktop/src/main/index.ts
+++ b/apps/desktop/src/main/index.ts
@@ -35,6 +35,8 @@ import {
   writeCliPrefs,
   type CliStatus,
 } from './cli-integration'
+import { acceleratorForCommand, appKeyboardCommands } from '../shared/keyboard-commands'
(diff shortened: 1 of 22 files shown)


ChrAlpha:
@codex review

sheepbox8646 (approved):

PR #579, opened 2026-06-06
feat: bot create progress stream

ChrAlpha:
## Summary

- Add SSE-based Bot creation progress for container setup.
- Keep creation progress alive across navigation via a shared Pinia store and dedicated progress route.
- Preserve created Bot state when post-create container/setup steps fail, surfacing setup errors as warnings instead of losing the Bot.
- Route SSE non-2xx responses through the shared SDK response/error interceptor path.
- Harden progress route lifecycle cleanup for delayed redirects under KeepAlive.

## Verification

- `go test -count=1 ./internal/bots ./internal/handlers`
- `vitest run --config vitest.config.ts --project web src/composables/api/sseClient.test.ts src/pages/bots/new-progress.test.ts src/composables/api/useBotCreateStream.test.ts src/store/bot-create-progress.test.ts`
- targeted ESLint passed
- `git diff --check`

PR #597, opened 2026-06-09
chore(spec): generate swag via go tool directive for reproducibility

Fodesu:
## Summary

The swagger `go:generate` directive used `swag@latest`, which is
non-deterministic — it ignores the `swag v1.16.6` pinned in `go.mod` and
resolves to whatever version the toolchain returns, so regenerating produces
spurious diffs that vary by machine/time.

Register `cmd/swag` as a go.mod `tool` dependency and switch the directive to
`go tool swag`, so generation always uses the version pinned in `go.mod`.
Future bumps only touch `go.mod` (`go get -tool ...`), never `swagger.go`.

Regenerating with the pinned v1.16.6 also syncs a stale spec: `GET /users/me`
declares `@Failure 401` in source (since #579) but the committed spec/SDK never
included it.

## Changes

- `internal/handlers/swagger.go`: `go run …swag@v1.16.6` → `go tool swag`.
- `go.mod` / `go.sum`: add the `tool github.com/swaggo/swag/cmd/swag` directive.
- `spec/{swagger.json,swagger.yaml,docs.go}`: regenerated; adds the missing
  `401` on `GET /users/me` and normalizes operation key ordering.
- `packages/sdk/src/types.gen.ts`: the corresponding `401` type.


Fodesu:
@codex review

chen-ran (approved):

PR #599, opened 2026-06-09
perf(web): stabilize chat stream rendering

ChrAlpha:
## Summary
- Preserve streamed assistant block identity by upserting blocks in place instead of replacing arrays per update.
- Reconcile refreshed chat turns in place and adopt server IDs onto optimistic turns, keeping just-sent messages mounted under stable client keys.
- Stop client-clock session restamping, remove orphaned chat stream components, and drop the legacy local SSE delta parser path.


ChrAlpha:
@codex review

sheepbox8646 (approved):

PR #603, opened 2026-06-09
feat(web/chat): refine process rail and background-task UX

ChrAlpha:
stacked on top of #599 

Before this PR, thinking, tool calls, and background-task updates were all rendered at the same visual weight as the final answer. Long tool runs made the chat hard to scan, live reasoning could feel noisy, short tool calls flashed past too quickly, and background tasks were easy to lose once their row scrolled off screen.

## major changes

- Moving thinking and tool calls into a _quieter process rail_, while keeping final answers full-width and easy to read.
- Showing live thinking as a calm one-line peek instead of expanding raw streamed reasoning by default.
- Rolling active tool calls through a single readable slot with dwell time, avoiding placeholder-only rows.
- _Collapsing_ settled thinking/tool runs into summary rows, while preserving expandable detail.
- Keeping live or interactive states visible: pending approval, pending user input, running/stalled background tasks are not folded away.
- Surfacing off-screen background tasks with a floating pill that can jump back to the relevant message.
- Respecting reduced-motion preferences and improving disclosure semantics with aria-expanded.

## boundaries

The PR is scoped to the chat UI layer. It does not change backend behavior, stream protocol shape, model invocation, persistence, or the already-merged render reconciliation/perf path.


ChrAlpha:
@codex review

sheepbox8646 (approved):

PR #716, opened 2026-06-27
feat(context): compaction summary parity for discuss and pipeline chat

ChrAlpha:
Slimmed restack of this PR after #856 consolidated the chat/thread domains. The previous revision implemented summary parity by duplicating composition across the flow and pipeline paths; after #856 both paths converge on `timeline.ComposeContext`, so this revision implements artifact-aware composition once and wires both call sites. The durable event-delivery machinery from the previous revision (delivery claims, leases, queued/injected replay, decision redelivery) is split out and will follow the Run Journal shape from the channel-boundary spec (§6) as a separate PR.

## Summary

- **Artifact-aware composition, one implementation.** `timeline.ComposeContextWithArtifacts` replaces covered rendered segments and covered turn responses with their compaction summary at the covered slot — rendered slot first, then response slot, then the durable anchor — preserving order within same-millisecond ties in both domains. Segments edited after an artifact's coverage snapshot stay visible; blank, coverage-malformed, and coverage-less (pre-coverage legacy) artifacts are never projected, so they can never swallow or duplicate history.
- **Both call sites share it.** The chat pipeline branch (`buildMessagesFromPipeline`) and the discuss trigger load the session's active artifact frontier via `compaction.TimelineArtifactSource` and compose through the same function; parity is structural rather than mirrored. Budget trimming pins summaries so they survive a dropped prefix.
- **Discuss turns trigger compaction.** Native discuss turns re-evaluate compaction pressure with the chat trigger policy (summary-excluded numerator in the chat estimator's unit, budget-clamped threshold). Artifact summaries are identified by an explicit `CompactionArtifactID` carried through `turn.DiscussMessage`, never by content sniffing. ACP discuss already inherits the chat trigger through `streamTurnChat`.
- **Mention gating.** Self-sent segments never wake the bot; compaction coverage never erases an unconsumed mention (gating is watermark-only); image extraction uses the coverage-filtered timeline so a summarized image is not re-attached.
- **Monotonic event cursors.** `bot_session_events` payloads carry a JSON-safe monotonic cursor from a sequence seeded above both the wall clock and trusted-capped legacy data. Discuss consumption tracks a dual watermark and a segment counts as consumed only where every informative domain agrees — a covered cursor alone never proves consumption (concurrent inbound workers can allocate against source order), and a watermark seeded from persisted replies alone carries no cursor to compare. Payload cursors are trusted only within `(0, MaxTrustedEventCursor]`; backup import strips instance-local cursors and zeroes restored watermarks, so restored history gates in the source-time domain and import never mutates the local sequence.

## Boundary

- No delivery claims/leases, queued or injected durability, response replay, or decision-event redelivery here — that machinery follows as a separate PR aligned with the Run Journal (spec §6). In-process dispatch semantics on main are unchanged.
- Migration `0120_discuss_event_cursor` (renumbered after main claimed 0115–0119) contains only the cursor sequence, the discuss cursor column, and their trusted-capped backfill.
- Compaction selection and artifact persistence are unchanged; this PR only projects the existing frontier into composition.
- Accepted tradeoffs: (1) a hostile pre-existing dataset that drives the migration floor to the trusted ceiling degrades later allocations to source-time gating, chosen over widening projection trust; (2) losing a durable watermark over a compacted consumed mention may re-wake a session once, chosen over erasing pending mentions; (3) an insert-failure delivery keeps its stamped cursor and may replay once after recovery.

## Verification

- `go build ./...`, `go test ./...`, `golangci-lint run ./...` — all clean.
- PostgreSQL migrate up → down → up on an ephemeral postgres:16; replayed against a seeded hostile dataset (oversized and non-numeric payload cursors) to confirm the migration neither aborts nor exhausts the sequence.
- Integration guards for 0120 (`internal/db/discuss_event_cursor_migration_integration_test.go`) run green against a live database, covering the schema round trip and the clock-seeded sequence floor.
- Review: 14 rounds of adversarial review (external reviewer) plus a four-lens internal review (concurrency, regressions vs base, schema/contract, test integrity). Each accepted finding was fixed test-first, and every new guard was mutation-verified — the implementation was reverted in a scratch copy to confirm the test actually fails.
- Exact head: `0e271333` — CI green (14 checks pass, publish job skipped as PR-only).

### DIFF
--- a/cmd/internal/channel/providers.go
+++ b/cmd/internal/channel/providers.go
@@ -92,10 +92,11 @@ func provideEventStore(log *slog.Logger, queries dbstore.Queries) *timeline.Even
 	return timeline.NewEventStore(log, queries)
 }
 
-func provideDiscussDriver(log *slog.Logger, eventStore *timeline.EventStore, msgService *message.DBService) *discuss.DiscussDriver {
+func provideDiscussDriver(log *slog.Logger, eventStore *timeline.EventStore, msgService *message.DBService, queries dbstore.Queries) *discuss.DiscussDriver {
 	return discuss.NewDiscussDriver(discuss.DiscussDriverDeps{
 		MessageService: msgService,
 		CursorStore:    eventStore,
+		Artifacts:      compaction.NewTimelineArtifactSource(queries),
 		Logger:         log,
 	})
 }
--- a/db/postgres/migrations/0001_init.down.sql
+++ b/db/postgres/migrations/0001_init.down.sql
@@ -41,6 +41,7 @@ DROP TABLE IF EXISTS bot_session_events CASCADE;
 DROP TABLE IF EXISTS bot_session_discuss_cursors CASCADE;
 DROP TABLE IF EXISTS bot_sessions CASCADE;
 DROP SEQUENCE IF EXISTS session_runtime_fencing_token_seq;
+DROP SEQUENCE IF EXISTS bot_session_event_cursor_seq;
 DROP TABLE IF EXISTS bot_channel_routes CASCADE;
 DROP TABLE IF EXISTS channel_identity_bind_codes CASCADE;
 DROP TABLE IF EXISTS bot_channel_configs CASCADE;
--- a/db/postgres/migrations/0001_init.up.sql
+++ b/db/postgres/migrations/0001_init.up.sql
@@ -554,6 +554,11 @@ CREATE INDEX IF NOT EXISTS idx_bot_channel_routes_bot ON bot_channel_routes(bot_
 -- bot_sessions: chat sessions within a bot, optionally linked to a channel route.
 CREATE SEQUENCE IF NOT EXISTS session_runtime_fencing_token_seq AS BIGINT NO CYCLE;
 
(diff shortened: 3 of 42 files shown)


ChrAlpha:
@codex review

ChrAlpha:
@codex review

ChrAlpha:
@codex review

ChrAlpha:
@codex review

ChrAlpha:
@codex review

ChrAlpha:
@codex review

ChrAlpha:
@codex review

ChrAlpha:
@codex review

ChrAlpha:
@codex review\n\nPlease review exact head `6e2911aa16e12d2a4d3a72f76bc7b0f9b0ffbe36`, including the queued route-handoff persistence fix and its deterministic regression test.

ChrAlpha:
@codex review

ChrAlpha:
@codex review

ChrAlpha:
@codex review

ChrAlpha:
@codex review

HoneyBBQ (approved):

PR #733, opened 2026-07-01
refactor(chat): per-session view architecture, chat stays single-copy

qqqqqf-q:
## Summary

Lays the groundwork for a per-session view architecture without shipping
user-facing chat split — chat remains one live pane at a time by design, but
the `SessionView` registry (per-session state + SSE + ref-counted lifecycle)
stays exercised for future non-chat split surfaces (e.g. a subagent rendering
pane).

- Keep the `SessionView` architecture live (per-session storage, SSE
  orchestration, ref-count open/release) but cut the premature background-send
  / paging plumbing that assumed a second live chat pane.
- Background session views persist across focus switches: releasing a view
  stops its dedicated SSE but keeps the view object so the bot-wide WebSocket
  keeps appending in-flight generation to it; views are only dropped on a real
  session/bot switch or auth reset.
- `splitGroup`/`canSplit` exclude chat panels — split stays available for
  every other panel type.
- Fixes a real regression the above surfaced: `initialize()`'s stream teardown
  was clearing `sessionViews` (shared with the background-persist path), which
  could wipe a session's view mid-flight during the async bot-switch watcher.

## Test plan

- [x] `pnpm --filter @memohai/web test src/store/chat-list.test.ts` — 106 tests pass
- [ ] Manual: switch bots/sessions while a generation is streaming in the
      background, confirm it keeps appending and isn't lost on refocus
- [ ] Manual: split a non-chat panel (terminal/browser/display), confirm chat
      panels still offer no split action

### DIFF
--- a/apps/web/src/pages/home/components/chat-pane.vue
+++ b/apps/web/src/pages/home/components/chat-pane.vue
@@ -1086,20 +1086,28 @@ const agentChanging = ref(false)
 const acpModelChanging = ref(false)
 const inputDrafts = useStorage<Record<string, string>>('chat-input-drafts', {})
 
+// props.sessionId is the pane's session (null = unsaved draft). The pane reads
+// the store's global focused-session state directly (single-pane: the pane's
+// session IS the focused one). paneSessionId is still tracked below solely to
+// hold the session's view ref via syncSessionViewHold — the per-session
+// ref-count lifecycle stays live for future split surfaces, but the reads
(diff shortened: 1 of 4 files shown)


qqqqqf-q:
关闭：base 已因 #719/#734（Turn DAG History 引入又撤销）产生大量语义冲突，收益不值得这次合并成本，搁置。

PR #746, opened 2026-07-06
fix(devenv): verify pnpm patched-dependency content + pin store inside node_modules volume

qqqqqf-q:
## Problem

The 18082 dev environment showed a broken dock tab strip: the `+` action pinned to the far right with a dead blank stretch after the tabs (screenshot symptom also reproducible on desktop when its cache was stale).

Root cause was **not** code: the dev container's `node_modules/.pnpm/dockview-core@7.0.2_patch_hash=…` directory had the **correct patch hash in its name but unpatched content**. The tab-strip CSS from #729 is written against the patched DOM (`+`/void moved inside `.dv-tabs-container`), so against the original DOM the tabs-container's `flex: 1 1 auto` swallowed the header and pushed the `+` to the far edge.

How it got poisoned: pnpm's store fell back to `<repo>/.pnpm-store` on the bind mount — the compose `pnpm_store` volume mounts at `/root/.local/share/pnpm/store`, a **different filesystem** from the `node_modules` volume, so pnpm can't hardlink from it and silently uses the repo-root fallback instead. That fallback store is shared between the host pnpm (10.29.3) and the container pnpm (10.34.4). One corrupted materialization of the patched package then got reused by every subsequent install, forever — pnpm trusts the hash in the directory name and never re-reads content.

Verified pnpm itself is fine: a clean minimal repro in the same container with the same patch file applies correctly, both fresh and headless (lockfile) installs.

## Fix

1. **`scripts/check-patched-deps.mjs`, wired as root `postinstall`** — for every `patchedDependencies` entry in `pnpm-workspace.yaml`, parse the .patch file and assert each patched file's most distinctive added line is actually present in the installed copy. Content is the ground truth; the directory name is not trusted. A poisoned install now fails loudly at install time with a recovery recipe, instead of surfacing as an unexplainable UI bug. Generic — automatically covers any future patch added to `patches/`.
2. **devenv compose (all 4 variants)** — set `npm_config_store_dir=/workspace/node_modules/.pnpm-store` for the `deps` service so the store lives **inside the node_modules volume**: same filesystem (hardlinks work), container-only (never shared with the host pnpm), and wiped together with the volume. Removed the now-unused `pnpm_store` named volume.

## Test plan

- [x] Positive: script passes against a healthy install (host + worktree `pnpm install` end-to-end, postinstall runs and passes)
- [x] Negative: reverting the patched lines in the installed dockview-core file makes the script fail with exit 1 and name the exact file; restoring makes it pass again
- [x] All 4 modified compose files pass `docker compose config -q`
- [x] eslint clean on the new script
- [ ] Fresh `mise run dev` on a machine without an existing `.pnpm-store` (store lands inside the node_modules volume)

## Notes for reviewers

- The checker is dependency-free by design (it runs as postinstall, i.e. it can't rely on the very node_modules it validates), hence the line-scanner for the YAML block instead of a YAML parser.
- Removal-only hunks are skipped: absence of a removed line can't be asserted safely (the text may legally occur elsewhere in the file). Every current patch file has added lines in every hunk section, so coverage is full today.


qqqqqf-q:
Self-review pass (0f61c4726) found and fixed four gaps in the first commit:

1. **P0 — production web image build was broken**: `docker/Dockerfile.web` never COPYs `scripts/`, so the new root postinstall failed `pnpm install` in the builder stage. Now copies exactly `scripts/check-patched-deps.mjs` before the install layer. Verified by simulating the Dockerfile's COPY set in a clean dir + full install: postinstall runs and passes.
2. **Recovery recipe was incomplete**: it omitted `apps/*/node_modules/.vite`. The dep optimizer's cache key is the lockfile, which cache poisoning does **not** change — a bundle built from the poisoned install would be served even after node_modules is fixed (this exact trap occurred in the original incident). The printed recipe now includes it, with the reason.
3. **`npm_config_store_dir` was only on the `deps` service**: a `pnpm install` run inside the `web` container (debugging, one-off upgrades) would still fall back to the host-shared repo-root store. All four compose files now pin the same store on `web` too.
4. **Peer-variant coverage**: the checker only inspected the first `_patch_hash` directory; peer variants materialize as separate copies and each must carry the patch. It now verifies every matching directory.

Also empirically verified what the first commit had only reasoned about:
- `pnpm store path` inside the container resolves to `/workspace/node_modules/.pnpm-store/v10` with the env var set;
- switching the store over an **existing** node_modules volume auto-purges and reinstalls under `CI=true` (which `deps` sets) — existing dev volumes migrate on the next `mise run dev` with no manual steps (without CI/TTY pnpm aborts with `ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY`, which is why `web` only pins the var and installs should keep going through `deps`);
- the in-volume store survives subsequent installs.

Known accepted limits: `pnpm install --ignore-scripts` skips the guard (all CI workflows use plain `--frozen-lockfile`, which does run it); removal-only patch hunks aren't asserted (absence of a line can't be checked safely — every current patch adds lines in every file it touches).

qqqqqf-q:
Third commit (e5b1bd371), prompted by the question *"if this merges and a colleague with a poisoned env pulls latest and reruns, is their dockview guaranteed patched?"* — the honest answer for the previous two commits was **no**. Walking that scenario end to end surfaced two remaining holes, both now closed:

**Hole 1 — stale dep-optimizer caches survive every repair.** Vite keys its cache on the lockfile, which cache poisoning does not change. Empirically verified that even pnpm's store-switch purge (triggered once by this PR's store relocation) leaves `apps/*/node_modules/.vite` intact — it lives on the bind mount, not the volume. A colleague whose cache was built during the poisoned era would pass the install guard yet keep being served pre-patch code. The script now also derives each patch's most distinctive **removed** line (code that must not exist post-patch) and deletes any optimizer cache whose chunks still contain it (whitespace-insensitive match; deletion is self-healing — vite re-optimizes on next start).

**Hole 2 — the pull-without-install path.** This PR doesn't change the lockfile, so a host-flow dev can pull it and restart vite without `pnpm install` ever running — postinstall never fires. `apps/web/vite.config.ts` now runs the same guard in `configResolved` (before the optimizer reads its cache): poisoned installs fail dev/build loudly with the recovery recipe, stale caches are swept in the same pass. This also covers `--ignore-scripts` installs, closing the caveat from the previous comment.

**Guarantee matrix after this commit** (colleague pulls merged main, then reruns):
| Path | What happens |
|---|---|
| `mise run dev` (container) | compose config changed → deps reinstalls; store relocation purges node_modules; postinstall verifies + sweeps caches; if anything is still wrong, deps exits non-zero and web never starts |
| host `vite` dev / `vite build`, no install | config file changed → running vite auto-restarts → plugin verifies + sweeps; poisoned = loud failure with recipe |
| prod image build | postinstall verifies during `pnpm install` (script is COPYed) |

Remaining honest limit: a long-running vite started **before** the pull that somehow doesn't restart on config change, and a dev who neither installs nor restarts anything — no hook can run before code does. Everything that "re-runs" hits a checkpoint.

Tested: healthy pass / planted poisoned chunk → cache removed with warning / healthy chunk kept / `vite optimize` runs the plugin and exits 1 on a poisoned install / eslint clean / full pre-commit Go+web hooks green.

qqqqqf-q:
追加 commit a5947ea11:apps/desktop 的 electron-vite renderer 也接入同款 configResolved 守卫。desktop renderer 直接 import web 模块(含 dockview),且有自己独立的 `apps/desktop/node_modules/.vite` 优化器缓存(实测 70MB),暴露面和 web 完全一致——之前只挂 web 是覆盖漏洞。

开销实测(担心阻塞正常开发路径的同学看这里):
- 健康安装下守卫全程 **~0.12s**;
- 最坏情况 sweep(web+desktop 共 150MB / 900 个缓存 js 全扫)**~0.54s**,且只在缓存存在时发生;
- 误删风险已排查:所有 patch 的 removed-line marker 均不与 added 行重合(无 moved-line 误报),即使误删也只是删缓存目录 → vite 下次启动自动 re-optimize;
- 唯一会**硬阻塞**的情况 = 安装内容确实被污染,此时阻塞正是目的,stderr 附一键恢复处方。

验证:desktop `electron-vite build` 日志中 renderer 阶段输出 `✓ patched dependencies verified against installed content`。

qqqqqf-q:
合并前做了一轮对抗性审查(独立 agent,逐行读脚本 + 全仓库搜 Dockerfile/CI + 对照 pnpm 上游源码验证 store 迁移行为)。结论:无 CRITICAL;1 HIGH + 4 MEDIUM,其中必修项已在 5cebd8f0c 落地:

- **H1 恢复处方漏了新 store**:处方原来只删仓库根 `.pnpm-store`(旧向量),但本 PR 把 store 钉进了 `node_modules/.pnpm-store`——毒的本体在 store 条目里,不删它重装只会重新硬链坏拷贝,处方在它为之而生的场景下会死循环。已补删,并给 Docker 镜像构建场景注明用 `docker builder prune`(BuildKit cache mount)。
- **M1 引号/无版本 key**:pnpm 对 scoped 包写 `'@scope/pkg@1.0.0':`(带引号),原解析不剥引号 → 第一个 scoped patch 落地当天所有安装全线误报失败;无版本 key(合法形态)同样产生垃圾前缀。已修并加合成测试验证(见下)。
- **M2 零 marker 静默放行**:patch 提取不出任何可验证新增行时,原实现仅凭目录存在就打 ✓,架空 exit 0 语义。现在拒绝认证并明确报错。
- **M3 sweep 并发韧性**:`.vite` 目录在 bind mount 上被宿主 vite / 容器 vite / desktop electron-vite 三方共享,扫描期间他方轮换临时文件或存在坏 symlink 会把守卫炸成裸 crash。sweep 现在逐条容错、整体异常降级为 warning(它是自愈步骤,不是认证的一部分;verify 的失败语义不变)。

验证:健康路径通过;合成 fake-repo 测试 5 组全过(带引号 scoped + 无版本 key 健康通过 / 投毒内容 exit 1 / 零 marker 拒绝认证 / patch 缺失清晰报错 / 坏 symlink 不崩且毒缓存照常清除)。

审查 LOW 项(删除文件型 patch section、verify marker 最短长度、旧 pnpm_store volume 孤儿、package.json 形态 patchedDependencies、vitest 独立入口)记录在案,不阻塞合并;旧 volume 清理建议:升级后可手动 `docker volume rm <project>_pnpm_store`,首次 `mise run dev` 时 pnpm 检测到 storeDir 变化会自动清空重装 node_modules volume(一次性,几分钟)。

PR #760, opened 2026-07-09
fix(web): dark-mode autofill readability on login page

qqqqqf-q:
## Summary
- Browser-native autofill forces its own background on `<input>` and only honors `-webkit-text-fill-color`, not `color` — with `--foreground` white text in dark mode, autofilled fields (browser-remembered credentials) became unreadable.
- Ports the inset-box-shadow override hack from the SaaS login page (`memohai/Memoh-SaaS`), where this was found and fixed first.
- The password field needed its own selector: `PasswordInput` renders through `InputGroupInput`, which overrides `data-slot` to `"input-group-control"` via Vue's attrs fallthrough — a selector scoped to `[data-slot="input"]` alone silently misses it. Both are covered now, with a comment explaining why.

## Related
Companion fix in the SaaS repo (draft, held open for reference — that repo isn't a fork of this one so the CSS doesn't sync automatically): memohai/Memoh-SaaS#TBD

## Test plan
- [x] `vue-tsc --noEmit` clean
- [x] `eslint` clean
- [ ] Manual check: trigger browser autofill (saved username/password) on `/login` in both light and dark mode, confirm text is readable

### DIFF
--- a/.agents/skills/memoh-web/SKILL.md
+++ b/.agents/skills/memoh-web/SKILL.md
@@ -618,6 +618,25 @@ These are judgment rules. They are the difference between "it renders" and "it's
 
 ### 1. Interrogate every line of copy
 
+**P0 — Label ≠ placeholder (never show the same noun twice).** A field label and its
+placeholder must not be the same string, and must not read as the same bare noun sitting
+twice in the viewport (e.g. label `Password` + placeholder `Password`, or label `Email` +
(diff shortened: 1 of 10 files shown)


qqqqqf-q:
Superseded — this bundled the full login-page control-language port (colors, hover, edges, primary fill), not just an autofill fix. Reopening under refactor/login-control-language-port with a title/description that reflects the actual scope.

PR #769, opened 2026-07-11
chore(sdk): upgrade Hey API to 0.99.0

qqqqqf-q:
## Summary

- upgrade `@hey-api/openapi-ts` from `0.92.3` to `0.99.0` and regenerate the checked-in TypeScript SDK
- add an SDK-only TypeScript gate to `mise run sdk-generate`
- adapt Web error paths to the new optional `request` / `response` contract
- add regressions for REST 401 handling, network failures, and the generated parameter template prototype-isolation fix
- exclude the display-prepare SSE POST from Pinia Colada mutation generation, while keeping the underlying SDK SSE operation

## Why

`0.92.3` is behind the current stable generator and belongs to the version range affected by `GHSA-hhx9-57xq-r5rw`. The current Memoh operation surface does not expose a confirmed exploit path through the affected helper, but upgrading removes the affected generator version and regenerates the template with null-prototype parameter slots.

Hey API 0.99 also models request-construction and network errors more accurately: an error result may not contain a `Response`. The affected Web callers now handle that state without throwing a secondary exception.

## Compatibility notes

- `@memohai/sdk/colada` intentionally no longer exports `postBotsByBotIdContainerDisplayPrepareMutation`. This is a 0.x breaking removal. The old generated helper destructured `{ data }` from an SSE `{ stream }` result, returned `undefined`, and never consumed the lazy generator, so it did not initiate the HTTP request. Memoh uses the stream-aware display wrapper instead. The underlying `postBotsByBotIdContainerDisplayPrepare` SDK operation and its types remain available.
- SDK generation now requires Node `>=22.18.0`, matching Hey API 0.99. The repository pins Node 25 through mise.

## Verification

- rebased onto current `origin/main` (`acf3b1f8e`)
- `mise run sdk-generate` passed and reproduced the generated SDK byte-for-byte
- `mise run lint` passed, including UI contract, ESLint, and Go lint
- repository pre-commit checks passed, including the full Go test suite
- `pnpm --filter @memohai/sdk typecheck` passed
- `pnpm --filter @memohai/desktop typecheck` passed
- `pnpm --filter @memohai/web build` passed after the rebase
- exact `docker/Dockerfile.web` image build passed after making the SDK tsconfig self-contained
- focused Web regressions passed, including auth, network-error, SSE, WebSocket, and parameter-isolation coverage
- production preview rendered the login page in Chromium with 0 console errors and 0 warnings
- `git diff --check` passed
- `pnpm audit` no longer reports the Hey API advisory
- latest GitHub CI passes Lint, Web/Server/workspace Docker builds on amd64 and arm64, and Desktop builds on macOS, Ubuntu, and Windows

## Existing baseline failures

After rebasing onto current main, full Vitest finishes with 68 files passing and 2 files failing. Neither failing path is changed by this PR:

- `apps/web/src/components/add-provider/index.test.ts`: the test mock omits `FORM_ITEM_INJECTION_KEY`
- `apps/web/src/pages/home/composables/useSubagentList.test.ts`: the test mock omits the newly required `ensureACPRuntime` export

`pnpm audit --prod` still reports 58 existing advisories unrelated to this upgrade, including 19 high-severity findings. Those should be handled in a separate dependency-governance change rather than mixed into this generator upgrade.


qqqqqf-q:
Follow-up from the first CI run: Docker Web failed because the new SDK tsconfig extended `/build/tsconfig.json`, which is intentionally outside `docker/Dockerfile.web` build inputs. Commit `4ac098d5e` makes the SDK typecheck config self-contained. I reproduced the exact Dockerfile locally after the fix; the full image build passed and Vite transformed all 7,216 modules without the missing-tsconfig error.

PR #790, opened 2026-07-14
test: add agent runtime contract harness

Fodesu:

PR #792, opened 2026-07-14
feat(runtime): add local session state core

Fodesu:
## Summary

- add the in-process session runtime manager and canonical snapshot/delta state model
- keep the memory backend local with session subscriptions, abort, and steer routing
- cover lifecycle, sequencing, replacement, and command behavior with focused Go tests

## Stack

Depends on #790.

## Verification

- `go test ./internal/sessionruntime`


Fodesu:
Closing and recreating this stack using fork-only branches; no upstream feature branches will be retained.

PR #800, opened 2026-07-14
feat(runtime): add session runtime foundation

Fodesu:
## Summary
- add the session runtime contract harness and memory runtime core
- add Redis/Valkey coordination and PostgreSQL fencing foundations
- keep frontend contracts limited to behavior available at this layer

## Validation
- `go test ./internal/config ./internal/sessionruntime ./internal/runtimefence ./internal/message ./internal/toolapproval ./internal/userinput ./internal/acpagent`
- `pnpm test --run --testTimeout=15000 apps/web/src/composables/api/useChat.ws.test.ts apps/web/src/store/chat-list.test.ts`
- `pnpm --filter @memohai/web exec vue-tsc --noEmit`

### DIFF
--- a/.github/workflows/eslint.yml
+++ b/.github/workflows/eslint.yml
@@ -20,6 +20,7 @@ on:
       - "pnpm-workspace.yaml"
       - "eslint.config.mjs"
       - "tsconfig.json"
+      - "vitest.config.ts"
       - ".github/workflows/eslint.yml"
   pull_request:
     branches: ["main"]
@@ -40,6 +41,7 @@ on:
       - "pnpm-workspace.yaml"
       - "eslint.config.mjs"
       - "tsconfig.json"
+      - "vitest.config.ts"
       - ".github/workflows/eslint.yml"
   workflow_dispatch:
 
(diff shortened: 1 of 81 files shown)


qqqqqf-q (approved):
Reviewed the architecture and ran the full test matrix locally against real Redis 7 + PostgreSQL 16 (memory contract suite, Redis contract suite — verified 0 skips, runtimefence integration, migration 0111 up/down/re-run path, and the fence integration tests in message/acpagent/userinput/toolapproval). All green.

Highlights from the review:
- Epoch + monotonic Seq delta protocol with gap detection and snapshot-fallback is sound; no silent event loss path (buffer overflow emits a dropped-marker that forces reconciliation).
- Fence validation and business writes share one transaction (`FOR NO KEY UPDATE` lock-as-validation) — no check-then-act window. Lock ordering (bot → session) is explicit and covered by dedicated lock-order tests.
- Legacy/unfenced rows (`NULL` token) are verifiably untouched by supersede logic, and the migration test covers upgrade-from-existing-data, idempotent re-run, and sequence non-regression.
- The schema/query doc shared with the team matches the actual diff — including the honest callout that `LockSessionDecisionSequence` serializes per-session decision creation on all paths (fenced or not) to fix the `MAX(short_id)+1` race. Verified the lock strengths in SQL (`FOR KEY SHARE` on bots, `FOR UPDATE` only on the session row during decision creation), so the blast radius is per-session and negligible.

Non-blocking notes, fine as follow-ups:
1. `message/service.go` carries two fixes unrelated to fencing (asset-creation failures now propagate instead of `Warn`, and `DeleteMessagesByX` → `ClearHistoryByX`) — worth a line in the PR description confirming they're intentional.
2. `Manager` has ~28 `m.distributed != nil` branches; memory mode re-inlines the distributed invariants rather than living behind the backend interface. A third backend would touch all those branch points. The shared contract-test suite compensates, but worth keeping in mind.
3. The 7 `StartRun*` variants all funnel into one path — an options struct would shrink the API surface for integrators.
4. Orphaned-run reaping is lazy (only on `Snapshot()` calls); a session nobody subscribes to stays `running` until state TTL. Acceptable, but curious if an active reaper is planned.
5. Every `HandleAgentEvent` round-trips the full snapshot JSON through a Redis WATCH/MULTI transaction with an unbounded retry loop — likely the main perf question once #801 wires live token streaming through this. Will raise it there.

Foundation-only as advertised: verified `Activate`/`NextSessionRuntimeFenceToken` have zero production call sites, so deploying this alone changes no behavior.

PR #801, opened 2026-07-14
feat(runtime): integrate distributed session runtime

Fodesu:
## Summary

Backend half of the session-runtime cutover. Builds on the merged foundation in
#800 and wires distributed ownership, decision routing, retry/edit replacement,
and the snapshot/delta contract the Web client in #802 consumes.

- canonical `runtime_snapshot` / `runtime_delta` for active and terminal runs
- one owning server per active run, with commands routed to it across processes
- approval and `ask_user` routed to the active owner, or a fenced continuation
- lease detects a dead owner; a PostgreSQL fencing token blocks stale writes
- legacy WebSocket frames kept for older Web clients
- stable public error codes; private causes stay in server logs

## Architecture

```mermaid
flowchart TB
    Handler["internal/handlers<br/>LocalChannelHandler (HTTP/WS)"]
    Runtime["internal/sessionruntime<br/>Manager · ownership · subscriptions"]
    Decision["internal/decisionruntime<br/>approval / ask_user routing"]
    Flow["internal/conversation/flow<br/>retry/edit prep · agent stream"]
    Fence["internal/runtimefence<br/>PostgreSQL write token"]
    Message["internal/message<br/>StartCanonicalTurn / AppendCanonicalTurn"]

    Handler --> Runtime
    Handler --> Decision
    Runtime --> Flow
    Decision --> Flow
    Flow -->|"each step, guarded"| Fence
    Fence --> Message
    Runtime -.snapshot/delta.-> Handler
    Message -.message_created.-> Handler
```

## Distributed ownership

A run is stateful in-process: the server that admits it holds the Go abort
channel, injection channel, and the live agent stream. That server is the
**owner**. State that must survive a process (status, messages, ownership) lives
in the backend; the live control does not.

- **Owner server.** Ownership is recorded as a `StreamRef{OwnerID, Generation}`
  and enforced by `Manager.ValidateRunOwnership` (`internal/sessionruntime/ownership.go`).
  Only the owner can drive the run, because only it holds the in-process channels.

- **Redis command channel.** The pub/sub primitive
  (`RedisBackend.PublishCommand` / `SubscribeCommands`, keyed per owner as
  `commands:<ownerID>`) was added in the merged foundation #800. This PR wires it
  up: `Manager.DispatchActiveCommand` runs a command locally when
  `ownerID == m.ownerID`, otherwise `dispatchRemoteCommand` publishes it to the
  owner and waits for the result (`internal/sessionruntime/routed_command.go`).
[...]

PR #802, opened 2026-07-14
feat(web): observe canonical session runtime state

Fodesu:
> Depends on #801, rebased on `3d5c1e9e6`. Review the range `3d5c1e9e6...HEAD`.
> GitHub shows the inherited #801 commits because both PRs target `main`.

## Summary

Cut the Web client over to canonical runtime state, and persist agent output at
**step granularity** so what the client observes always matches durable history.

- shared SDK reducer validates epoch + sequence, else requests a resync
- chat store projects runtime state into assistant streams and the transcript
- each completed agent step is appended to canonical history under the fence,
  so mid-run reconnect, retry/edit, and terminal recovery see the same rows
- fixes retry/edit projection: old assistant removed, one user turn, new ID

## Layers

```mermaid
flowchart LR
    WS["useChat.ws.ts"] --> SDK["packages/sdk/session-runtime<br/>parse + reduce"]
    SDK --> Store["chat-list.ts<br/>runtime projection"]
    Store --> Streams["chat/assistant-streams.ts"]
    Store --> Transcript["chat/transcript.ts"]
    Transcript --> UI["chat-pane / message-item"]
    UI -.commands.-> WS
```

## Step-granular persistence + projection

```mermaid
sequenceDiagram
    participant UI as Web (chat-list)
    participant RT as sessionruntime
    participant Flow as conversation/flow
    participant DB as message (canonical turn)

    UI->>RT: send / retry / edit
    Flow->>DB: StartCanonicalTurn (persist user request)
    Flow-->>RT: EventHistoryCommit (history_committed, request id)
    RT-->>UI: delta → bind canonical request turn

    loop each completed agent step
        Flow->>DB: AppendCanonicalTurn (fenced write)
        Flow-->>RT: step messages
        RT-->>UI: runtime_delta (append, seq+1)
    end

    Flow->>DB: appendTerminalSnapshot (remaining messages)
    Flow-->>RT: terminal (canonical_ready, assistant id)
    RT-->>UI: delta → drop optimism / REST-reconcile
```

The client never waits for the whole turn to land: it binds the canonical
request turn on `history_committed`, streams each persisted step, and only
clears optimistic turns once terminal `canonical_ready` and the assistant ID
line up with the DB.

## Retry / edit

On admission the client keeps the optimistic stream and leaves the transcript
alone until `history_committed`. Then `replaceTailFromTurn()` drops the old
assistant, rebinds the surviving user row to the canonical `request_user_turn.id`,
and replaces (not duplicates) any hydrated assistant at that position.

## Validation

- `go test ./...`
- Web Vitest: 1045 passed
- `pnpm --filter @memohai/web typecheck` · `pnpm --filter @memohai/sdk typecheck`


qqqqqf-q (commented):
Reviewed the full cutover: ran the frontend suites (store + api: 25 files / 643 tests green; the six runtime-touched files re-run explicitly: 363 tests green), and also did a live end-to-end test of this branch deployed with the Redis backend — mid-run page refresh and cross-browser recovery both work; the canonical snapshot lands in Redis with epoch/seq and full run content as designed.

**The protocol layer is genuinely excellent.** Verified every recovery path: seq continuity checks are real (not ceremonial), gaps/epoch changes/dropped markers all funnel into resync, the awaiting-checkpoint barrier plus server-side ordering give a sound snapshot-then-delta guarantee, and there is no silent event-loss path. The reducer is a true pure state machine, the contract fixtures are genuinely shared three ways (generated from Go wire types, byte-level drift check, replayed through the real reducer, husky hook), and removing the send queue so reconnects can't replay stale commands is exactly right.

**One real issue — abort has no client-side watchdog.**

After this PR, abort only sends the frame and marks `abortRequested`; the stream settles solely when the runtime terminal state projects back. But there are paths where the terminal event never arrives — e.g. the server dies inside the soft-cleanup window (runner blocked on a ctx-unaware call until lease TTL), or any bug in the terminal-event chain. In those cases the session's loading state hangs forever. The old code couldn't get stuck (immediate local reject); the new code treats "the loop always closes" as an axiom. Suggest a client watchdog after abort (~15–30s): on timeout without a terminal state, settle locally and trigger a resync.

**Ask: CI path-filter regression.** This PR removes `apps/web/src/store/__fixtures__/runtime-*.contract.json` from the paths filters in go-ci.yml and eslint.yml (two places each). A fixture-only diff — which can happen when this stack is rebased/split — will now trigger no CI at all: neither the Go drift check nor the consumer tests. Please restore those paths.

**Non-blocking notes:**

1. `fetchSessionRuntime` (useChat.message-api.ts:65) is production dead code, only referenced by negative test assertions — delete it or add a comment explaining it's a sentinel for a future PR.
2. Structure suggestion (non-blocking): the runtime projection block (~1,200 lines) and the subscription/recovery block (~290 lines) in chat-list.ts are two self-contained controllers. I measured their coupling: all references are explicit — the subscription block touches only 12 externals (10 of which are its own bookkeeping state that can move with it), and the projection block's 26 externals are mostly transcript-controller methods already injected via deps. Extraction into `store/chat/` files following the `createTranscriptController` precedent would be mechanical, no logic change. Happy to do this as a follow-up PR ourselves if you'd rather keep this stack moving.
3. Two minor observations, no action needed: the delta path double-writes content (reducer and the projection layer each maintain append logic for the same text with no cross-check — healed by the next snapshot, but worth knowing); and retry/edit loses its optimistic replacement on legacy servers (harmless in same-version deployments).

internal/sessionruntime/abort.go
```diff
@@ -0,0 +1,276 @@
...
+	if !acknowledged {
+		return false, ErrCommandTargetNotActive
+	}
+	ctrl.stopCommands()
+	select {
+	case ctrl.abortCh <- struct{}{}:
+	default:
+	}
+	if ctrl.cancel != nil {
+		ctrl.cancel()
```
wishworldbetter, line 146: 这里可能永久卡住 session：abort 只 cancel runner，但 lease renewal 使用独立生命周期。如果下游调用不响应 context，run 会一直处于 aborting 并持续续租，前端 watchdog 也只会重试 abort。建议增加服务端 abort grace timeout，超时后撤销 owner、停止续租并标记为 lost/forced_aborted，同时补一个 runner 忽略 context 的测试。

qqqqqf-q:
## QA matrix doc pushed to this branch

Commit: `6c1125ec1` — `docs(qa): add PR802 session runtime QA matrix`

[`docs/qa/pr802-session-runtime-qa.md`](https://github.com/Fodesu/Memoh/blob/feat/session-runtime-web-cutover/docs/qa/pr802-session-runtime-qa.md)

Already confirmed Fail (evidence in chat):
- **BUG-1 P0** Ask user Submit → `The runtime command could not be completed.` / `user input request not found`
- **BUG-2 P1** Tool loop + multiple ⌘R → UI message duplication (API `format=ui` often single copy)

Rest of the matrix is still untested.

PR #818, opened 2026-07-16
feat(web): restyle ask_user prompt as a composer-style selector

sheepbox8646:
## Summary

UX polish for the `ask_user` tool prompt in the web chat. The pending Q&A used to render as a small white card floating above the composer, with its own visual language (card border, tiny text, footer Submit/Cancel row) that clashed with the input capsule below it.

The prompt is now a selector that **takes the composer's place** and wears its exact language — same `chat-composer-edge` hairline, `bg-surface-composer` fill, 20px radius and inner padding — so answering a question feels like part of the input surface rather than a bolted-on dialog:

- **Option rows** — full-width ghost rows with radio/checkbox indicators (selection shows via the indicator, not a persistent row fill).
- **Divider**, then a **"Tell the bot more…" input** for the free-text / custom answer; Enter submits directly.
- **Submit / Cancel pair at the bottom.** Cancel resolves the request as canceled and hands the focused composer back.

## Behavior details

- While a request is pending the textarea hides; picking an option reveals it again so the user can keep typing.
- For `single_select`, typing a custom answer clears the picked option and vice-versa, mirroring the backend's "exactly one of option / custom text" rule — Submit can never be armed with an invalid combination.
- A lone `text` question routes its answer through the same footer input; multi-question requests keep per-question inline inputs (one footer input can't answer several text questions).
- The bottom backdrop mask now measures whichever box stands in for the composer (shared `COMPOSER_MASK_BELOW_PX`), so no message bleed behind the capsule's bottom corners.
- i18n: new `chat.tools.userInputTellMore` key in en/zh/ja.

## Verification

- `vue-tsc --noEmit` clean, ESLint clean, `check-ui-contract.mjs` passes (no new violations).
- `vitest run apps/web/src/store` — 444/444 pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/web/src/i18n/locales/en.json
+++ b/apps/web/src/i18n/locales/en.json
@@ -697,6 +697,7 @@
       "userInputExpired": "expired",
       "userInputPlaceholder": "Answer",
       "userInputCustomOption": "Other…",
+      "userInputTellMore": "Tell the bot more…",
       "submitUserInput": "Submit",
       "cancelUserInput": "Cancel",
       "userInputConnectionLost": "Connection lost. Reconnect and try again.",
--- a/apps/web/src/i18n/locales/ja.json
+++ b/apps/web/src/i18n/locales/ja.json
@@ -666,6 +666,7 @@
       "userInputExpired": "期限切れ",
       "userInputPlaceholder": "答え",
       "userInputCustomOption": "他の…",
+      "userInputTellMore": "Bot にもっと伝える…",
       "submitUserInput": "送信",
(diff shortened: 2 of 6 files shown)


qqqqqf-q (approved):
LGTM — ask_user 选择器和 composer 视觉统一方向没问题，diff 范围也干净（6 个 web 文件，无 ACP 混入）。底部遮罩跟着替身盒子走这点处理得细致。先合。

PR #863, opened 2026-07-26
chore(runtime): remove Kata Containers support

chen-ran:
Kata Containers support is no longer needed, so we’re removing it entirely to simplify future maintenance.


internal/container/containerd/service.go
```diff
@@ -31,13 +31,14 @@ import (
 
 var ErrTaskStopTimeout = errors.New("timeout waiting for task to stop")
 
+const runtimeType = "io.containerd.runc.v2"
```
xhebox, line 34: The old `DefaultContainerdRuntimeType` is a better and clearer name, near the definition of other constants. It would be best that all literal `runc.v2` can be replaced by a constant, which could just be another PR.

xhebox (commented):
LGTM

HoneyBBQ (approved):

PR #881, opened 2026-07-29
feat(agent): cache workspace context snapshots

sheepbox8646:
## Summary

- Materialize workspace-owned Agent context (Hooks, Skills, `AGENTS.md`, `MEMORY.md`, `PROFILES.md`, and `HEARTBEAT.md`) in PostgreSQL per Bot and workspace target.
- Reuse one request-scoped snapshot for Agent preflight and resolve workspace descriptors without opening the filesystem/runtime client.
- Refresh snapshots when relevant file, Skill, Plugin, container lifecycle, import, terminal, or command mutations can change the source data.
- Make the Session Runtime the sole owner of the inject channel close, fixing a double-close panic exposed by live API testing.

## Why

Ordinary model-only turns previously resolved a live workspace and reread workspace context before the model call. In hosted deployments this can add E2B/runtime latency even when the Agent never invokes a workspace tool. The new generation-checked snapshot keeps source-of-truth files in the workspace while moving the hot read path to PostgreSQL; the runtime is opened only for initial/invalidated refreshes or actual workspace tools.

## Implementation notes

- Adds migration `0123_workspace_context_snapshots` with team RLS, target isolation, source status, content hashes, and requested/applied generation CAS.
- Coalesces refreshes in-process and safely handles superseding refreshes across processes.
- Fails closed for invalid user Hook configuration and retries stale/failed generations on the next request.
- Keeps background invalidation non-blocking where the originating operation is asynchronous.

## Validation

- `go test ./...`
- `CI=true golangci-lint run ./...` with the repository-pinned v2.10.1 binary
- `go test -tags integration ./internal/db -run 'TestWorkspaceContextMigrationRoundTrip|TestCanonicalInitContainsFinalTeamMembershipSchema|TestTeamRLSForceEnabled' -count=1 -v`
- `git diff --check`
- Restarted the Docker development server from this worktree; migration 123 applied and `/health` remains healthy.
- Real REST chat: hot-cache request accepted with HTTP 200, run completed, assistant returned `CACHE-OK`, and snapshot generation/hash remained unchanged with no refresh log.
- Real file API: modifying `/data/AGENTS.md` advanced the snapshot from `1/1` to `2/2` and changed its hash; restoring the exact revision advanced it to `3/3` and restored both the original file SHA-256 and snapshot hash.


sheepbox8646:
@codex review

chen-ran:
Necessary a discussion

chen-ran:
Wrote up an alternative design in #901 (docs only, draft) after reading through the code paths this touches.

The request-scoped snapshot here is the right idea and #901 keeps it. The disagreement is only about where the cache lands: `bot_workspace_context_snapshots` puts a per-turn write on the hot path, so context-cache write volume ends up tracking Bot chat volume and the hottest Bots make that table a write hotspot. #901 moves the hot path to Redis/Valkey (memory backend for single-instance OSS) with zero PostgreSQL writes per turn.

Two things I ran into while reading that seem worth flagging regardless of which design wins, since both are silent:

1. `skills.ListWithPluginRoots` calls `writeIndex` on every read, and the file it writes (`/data/.memoh/skills/index.json`) sits under `IndexDirPath`, which is itself a registered discovery root. So any fingerprint over discovery roots sees a change every turn and never hits. Failure mode is a log line saying sources changed, which reads as normal.

2. `bot_remote_runtime_bindings` is `UNIQUE (bot_id, runtime_id)`, so two Bots can bind one Remote Runtime and read the same physical `AGENTS.md`. Keying invalidation on `bot_id` means a write through Bot A leaves Bot B clean and serving stale context. On a Remote Runtime the user can also edit files directly without Memoh in the path.

Also: the inject-channel double-close panic fix in this PR is unrelated to caching and would be good to land on its own, ahead of whatever happens with the cache.

PR #891, opened 2026-07-29
feat(compaction): unify manual settings into the model-relative controller

ChrAlpha:
> **Stacked on top of #717**

## What this delivers

One compaction controller for every bot. #717 shipped the model-relative default policy behind a hard regime switch: `compaction_threshold == 0` selected the 50/75/40 controller, while any positive threshold fell back to the legacy absolute engine (one-shot async, ratio-driven target, no drain, separate gating). This PR removes the second regime so manual configuration parameterizes the same controller that zero-config runs.

- **Manual settings become parameter overrides, not a mode switch.** A positive `compaction_threshold` now only overrides the async trigger level (`min(threshold, 75% of the window budget)`); the hard backstop, keep target, bounded 3-pass drain, and quality gates stay identical to zero-config. `0` remains "no override".
- **`compaction_target_percent` replaces `compaction_ratio`.** Nullable, 1–99; `NULL` means the default 40% keep-share; it drives both the sync and async targets. Sending `0` (or any out-of-range value) clears the override back to the default — the same explicit-set semantics as the #717 compaction-model override. The engine's internal ratio mechanism survives only as the manual `/compact` full-compaction pin.
- **One blocking gate.** The caller-side 70% pre-gate and the legacy "always true" sync gate collapse into the single 75% hard threshold; percent-derived levels clamp to at least 1 token so small positive windows never read as "off".
- **Migration `0124`** adds the column, backfills `100 − compaction_ratio` for bots that were in the legacy regime (preserving their sync keep-share), drops `compaction_ratio`, and suspends/restores FORCE RLS around the backfill so owner-run migrations pass; `0001` updated as the canonical schema.
- **Web UI**: the ratio slider is gone; a keep-target input validates 1–99 (empty = default 40%), and the threshold copy documents `0` = automatic at 50% of the model window (en/zh/ja).
- **Backup compatibility**: importing an archive from before this change maps its stored ratio onto a target for legacy-regime bots, mirroring the migration.

## Verification

- `go build ./...` plus the application/settings/handlers/command/botbackup unit suites.
- Real-Postgres integration: the `0124` up/re-up/down/re-down/up cycle including non-superuser FORCE-RLS paths, explicit-set upsert semantics, and new-bot defaults.
- sqlc/swagger/SDK regenerated with a clean `git diff --exit-code` afterwards; vitest, targeted eslint, and the UI-contract check pass.
- The full `mise run lint` wrapper is blocked locally by pnpm's minimum-release-age policy on three pre-existing lockfile entries (the lockfile is untouched here); direct golangci-lint and eslint runs pass.

## Deliberately deferred

Provider-envelope budgeting and trimming unification stay context-restack scope; no measurement/numerator changes; no soft/hard percent knobs (the absolute threshold override plus the keep target cover the expressed intents without inviting gate-ordering foot-guns).


chen-ran:
There is one edge case worth tightening up. The hard backstop runs at 75% of the context window, but compaction_target_percent allows values up to 99. If it is set to 90, the sync path triggers at 75%, sees that the context is already below the 90% target, and does nothing.

ChrAlpha:
Good catch — fixed in a46214cb. The blocking backstop now caps its effective target at the soft share: `sync TargetTokens = min(compaction_target_percent, 50%)` of the window budget, so a pass that fires at 75% always makes progress and lands the session back under the soft trigger (restoring the 50/75 hysteresis). A target of 90 still applies unchanged to the async maintenance path, where sitting above the trigger just means maintenance waits until there is more than the keep-share to shed — that noop is benign and logged. Regression table covers target 90 (sync capped to 50%, async honored), the default 40, and sub-cap values passing through untouched.

chen-ran (approved):

PR #902, opened 2026-07-31
feat(agent): improve session title generation prompt and token budget

qqqqqf-q:
## Summary

Session auto-titles in the Recents sidebar were low quality: the old prompt only asked for "a concise title (max 30 characters)", so models paraphrased the user's first message verbatim — trailing question marks, request verbs, original word order — making the sidebar read like a chat log instead of a title list.

This PR replaces the prompt with a **rules + worked-examples** version and pins the completion token budget.

**The new prompt** (constant `titleGenerationPrompt`) encodes the target style explicitly:
- Noun phrase naming the **topic**, not a summary of the request (drop request verbs, no trailing "?")
- Per-language length bands (Chinese 6–10 preferred, 4–12 hard; English 2–6 words)
- Keep concrete proper nouns and vivid user wording; don't sand casual rants into officialese
- "X与Y" form for two-entity topics; a closed suffix list (分析/解析/介绍/建议/疑问/…) when one is needed
- Fixed fallback ("新对话") for meaningless input instead of hallucinating

**The token budget** (`titleGenerateMaxTokens = 4096`): reasoning models burn the completion budget on hidden thinking before answering; with provider/server default caps this can truncate the response to empty, and since the fallback title is already persisted by then, the session is stuck on the raw-prompt fallback forever. 4096 leaves ample headroom (a title is ~20 tokens).

## How the prompt was tuned

Offline A/B experiment against real first user messages (technical questions, life topics, pasted group-chat logs, degenerate input like "1"), run on small reasoning models:

| Input (abridged) | Old prompt | New prompt |
|---|---|---|
| "为啥 linter 这么费性能…" | "CI/Linter 为何如此吃性能？" | "Linter性能开销分析" |
| "下雨了，我的小电驴会有事吗" | "小电驴淋雨要紧吗" | "小电驴淋雨影响" |
| long Chinese group-chat paste | "Discussion on PR #717 and Harness Progress" (wrong language) | "Memoh harness 推进缓慢讨论" |
| "1" | "Math Problem Help" (hallucinated) | "新对话" |

Key experiment findings baked into the prompt's shape:
- Bare rules are not enough for small models — worked examples carry the style, but example **inputs** must stay disjoint from real traffic or the model copies them (overfitting).
- English instructions + "same language as the user's message" outperformed a fully-Chinese instruction version (which produced a bilingual format violation).
- Reasoning models with thinking enabled occasionally exhaust small token caps before emitting any text — hence the explicit budget.

## Known limitations / follow-ups (not in this PR)

- **No retry on failure**: if title generation fails once (LLM error, empty result), the session already holds the prompt-derived fallback title and `shouldGenerateSessionTitle` never allows a second attempt. Sessions that failed before this fix keep their raw-prompt titles. A retry mechanism (e.g. marking fallback titles as regenerable) is a separate, deliberate behavior change.
- Small models can still over-abstract on unusual inputs; the experiment showed this is a model-capability ceiling, not a prompt gap.

## Test plan

- [x] `go build`, `go vet`, package tests + pre-commit full Go lint/test suite — all green
- [x] Prompt validated offline across 12 diverse inputs × multiple runs (see table)
- [ ] Human QA: configure a title model (Settings → Profile → Title Model), start new sessions with varied first messages, confirm sidebar titles match the new style and that previously-working setups are unaffected

### DIFF
--- a/internal/agent/application/service_title.go
+++ b/internal/agent/application/service_title.go
@@ -24,13 +24,14 @@ import (
 const (
 	titlePromptMaxInputChars = 500
 	titleGenerateTimeout     = 60 * time.Second
-	// titleGenerateMaxTokens pins the completion budget. Without an explicit
-	// cap the budget falls back to provider/server defaults; reasoning models
-	// burn that budget on hidden thinking before answering and can return an
-	// empty title on long inputs (observed with deepseek-v4 thinking enabled,
-	// which the endpoint turns on server-side). 4096 leaves ample headroom —
-	// the title itself is ~20 tokens.
-	titleGenerateMaxTokens = 4096
+	// titleGenerateMaxTokens caps the title completion. Reasoning models burn
+	// budget on hidden thinking before answering, so the cap must clear their
+	// thinking plus a short title (provider defaults can be far smaller, which
+	// truncates the response to an empty title). But it must also leave room
+	// for the prompt on short-context chat models — provider templates still
+	// catalog 4097-token ones — or the request is rejected outright. 2048
+	// sits between the two failure modes.
+	titleGenerateMaxTokens = 2048
 )
 
 // titleGenerationPrompt produces short sidebar titles: a noun phrase naming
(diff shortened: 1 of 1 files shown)


qqqqqf-q:
Both P2 findings addressed in 41494ef05:

1. **Budget vs context window**: confirmed — `conf/providers/azure-openai.yaml` catalogs `gpt-3.5-turbo`/`gpt-35-turbo` with `context_window: 4097`. The budget is now `min(4096, context_window/2)` with a 256 floor (`titleCompletionBudget`), plus unit tests. Half the window leaves ample room for the title-sized prompt while keeping headroom for reasoning models' hidden thinking.

2. **Chinese-only rules scoped**: the `X与Y` form and the suffix list are now explicitly Chinese-only, with English forms given (`X and Y`, plain noun phrases). Re-ran the offline validation set (14 cases, incl. new English inputs) on the updated prompt: English inputs now produce e.g. "CORS Headers in Nginx" / "TCP vs UDP for Game Servers", Chinese cases show no regression, 14/14 succeeded.

Kept as-is: the meaningless-input fallback stays the fixed string 新对话 — a degenerate message (a bare "1", a sticker) carries no language signal to pick a locale from, and the frontend treats it as an ordinary title either way.

qqqqqf-q:
Simplified the budget fix in e356e4c3c per reviewer feedback: dropped the context-window clamp in favor of a fixed 2048 budget. 2048 leaves room for the prompt on 4097-token models (fixing the rejection issue) while still clearing a reasoning model's hidden thinking plus a short title. The clamp function and its test are gone — one constant does the job.

PR #924, opened 2026-08-01
feat(desktop): add native translucent sidebar

chen-ran:
## Summary

- use Electron's native macOS sidebar material instead of a CSS-only blur
- expose the material only through Desktop-specific surface hooks while keeping Web, Windows, and Linux opaque
- synchronize the renderer theme preference with Electron `nativeTheme`
- keep the dark reading surface at `#181818` and apply a 46% semantic sidebar tint so the native material remains visible
- prevent the persistent chat workspace from bleeding through the settings sidebar

## User impact

Memoh Desktop now has a calmer, Codex-like translucent sidebar on macOS. The native material follows the active window and system appearance, including when the window is moved between displays. Other platforms and the browser UI retain their existing appearance.

## Validation

- `pnpm --filter @memohai/desktop typecheck`
- `pnpm exec vitest run apps/desktop/src/main/window-chrome.test.ts apps/desktop/src/shared/theme.test.ts` (6 tests)
- focused ESLint on all changed TypeScript and Vue files
- `node scripts/check-ui-contract.mjs` (passed; 3 pre-existing warnings)
- `pnpm --filter @memohai/desktop build:dir`
- repository pre-commit Go lint/test and staged-file ESLint checks
- manually verified light and dark rendering on a macOS secondary display

### DIFF
--- a/apps/desktop/src/main/index.ts
+++ b/apps/desktop/src/main/index.ts
@@ -5,6 +5,7 @@ import {
   BrowserWindow,
   ipcMain,
   nativeImage,
+  nativeTheme,
   safeStorage,
   Tray,
   screen,
@@ -21,6 +22,7 @@ import trayIconPng from '../../resources/tray-icon.png?asset'
 import { acceleratorForCommand, appKeyboardCommands, type AppKeyboardCommand } from '../shared/keyboard-commands'
 import { dispatchFocusedWindowCommand } from './window-commands'
 import { dispatchRendererNavigate } from './window-navigation'
+import { macWindowChromeOptions } from './window-chrome'
 import { maybeSelfInstallMacOS } from './self-install'
 import { DesktopRemoteRuntimeManager } from './remote-runtime'
 import { isTrustedRendererUrl } from './renderer-trust'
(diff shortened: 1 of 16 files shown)


qqqqqf-q:
Added 02318689f on top of the original commit after evaluating the branch locally:

- **Tint 46% → 70%.** A vibrancy material's rendered color is hostage to whatever sits behind the window — clean over the default macOS wallpaper, muddy over a dark one. 70% theme color bounds that swing to 30%, keeping the sidebar recognizable across backdrops.
- **New sidebar ↔ chat seam** (native-material contexts only): 1px alpha stroke (light: black 6%, dark: white 7%) plus a soft shadow spilling left onto the sidebar (light: black 4%, dark: black 12%), so the chat plane reads as floating above the glass layer. Raw `--workspace-divider-color` washes out over the material, and a color-mix derivation read as the wrong hue — the alpha stroke is what survived tuning.
- **Material stays `sidebar`.** A/B'd against `menu` at 70% tint; the difference is not perceptible, and `sidebar` is the semantically correct material.

All of it is scoped to `data-native-sidebar-material` — Web, Windows, and Linux are untouched.

PR #966, opened 2026-08-06
refactor(models): give "off" one name across declarations and settings

HoneyBBQ:
## Problem

"Off" has two names. `disable` is what a user picks and what a bot stores; `none` is what a model may advertise in `reasoning_efforts`. Both survive into the picker, so a model advertising `none` renders two options — different labels, identical request.

The split is deliberate today. #563 introduced `REASONING_EFFORT_DISABLE` as the picker's off token — prepended unconditionally — and in the same change made `none` declarable. #926 then dropped the `reasoning_enabled` flag and promoted `ReasoningEffortDisable` into `internal/models` as "the settings/override value that says send no reasoning at all, which is why IsValidReasoningEffort rejects it", keeping `none` as the declarable spelling.

This PR argues that split costs more than it buys:

- **One state, two selectable options.** `KNOWN_EFFORTS` contains `none` and `availableEffortsForMode` prepends `disable` unconditionally, so a model advertising `none` offers both. Picking either produces the same request.
- **`none` sits in the tier ordering.** `orderedReasoningEfforts` lists it as the weakest tier, so `NearestEffortToMedium` can return it — an *active* reasoning config resolving to "no reasoning". #926 added a test guarding exactly this hazard for the `disable` sentinel; `none` walks through the same door because it is spelled like a tier.
- **Off is offered where it cannot be delivered.** The unconditional prepend shows the control on models that never declared they can be turned off, where the adaptor has no off shape to send and the model's default thinking behavior stands regardless.

Keeping two names is what makes all three possible: as long as off is spelled one way in settings and another way in declarations, nothing stops the declaration spelling from being treated as a tier.

## Change

`disable` becomes the single name for off on our side — both what a user picks and what a model advertises. `none` steps back to what it actually is: OpenAI's wire spelling, produced by the provider adaptor, exactly as the Anthropic path produces `thinking:{type:"disabled"}` for the same intent.

| Area | Change |
|---|---|
| `internal/models/types.go` | `validReasoningEfforts` accepts `disable`, drops `none`; `IsReasoningDisabled` still accepts `none` so values stored under the old spelling keep working |
| `internal/capabilities/discover.go` | LiteLLM's `supports_none_reasoning_effort` now lands on `disable` — the flag is named after OpenAI's wire value, but what it reports is that the model can be turned off |
| `internal/models/sdk.go`, `internal/agent/application/service.go` | off is translated to `none` at the OpenAI wire, and to nothing at all when the model cannot be turned off |
| `internal/command/reasoning.go` | `/reasoning` offers `off` instead of the `none` tier |
| `apps/web/.../reasoning-effort.ts` | `KNOWN_EFFORTS` holds only active tiers; declarable values are those plus `disable`; off is hoisted to the front only when the model advertises it |
| `conf/providers/*.yaml` | regenerated (`go run ./cmd/synccaps`) — 27 entries across 4 templates |

Why `disable` rather than `none` as the canonical value: `none` is OpenAI's vocabulary, and OpenAI has already changed it once — `minimal` was gpt-5.0's weakest tier and gpt-5.1 replaced it with `none`. Making it the neutral value binds the model layer to one vendor's naming. It also keeps the tier ordering honest: `disable` is not a tier, so it stays out of `KNOWN_EFFORTS` and the nearest-tier fallback can never resolve an *active* config to off.

## Behavior change worth calling out

Off is no longer offered on a model that does not advertise it. Previously the option appeared everywhere and, when picked on such a model, resolved to omitting `reasoning_effort` entirely — so the model's default thinking behavior stood and the control silently did nothing. Removing it makes the picker reflect what the model can actually do.

Existing bots whose stored effort is the legacy `none` keep working: `IsReasoningDisabled` still reads it as off.

## A hazard the change itself introduces

Advertising "off" puts the disable token into the effort level list, and `pickEffort` reads the bot's stored effort directly rather than through the tier ordering that deliberately excludes it. A bot parked on off, overridden for one message with a tier the model does not advertise, would hand `disable` back as the *active* tier and send it upstream, where no vendor knows the word. The requested effort was already guarded against exactly this; the second commit extends the same rule to the stored one.

This is the same hazard `orderedReasoningEfforts` and `NearestEffortToMedium` are shaped to avoid — worth noting that `pickEffort` bypasses both by reading settings straight.

## Verification

- `go build ./...`; `go test` on `internal/models`, `internal/capabilities`, `internal/agent/application`, `internal/command`
- `go run ./cmd/synccaps -check` clean; it correctly reports `templates are stale` if a `none` entry is put back, so the templates and the discovery mapping stay in agreement
- `vitest run src/pages/bots/components/reasoning-effort.test.ts` — 13 passing

New assertions cover the parts that carry the intent: `disable` is declarable and `none` is not, the picker offers off exactly once and only when advertised, and the nearest-tier fallback never returns off.

### DIFF
--- a/apps/web/src/i18n/locales/en.json
+++ b/apps/web/src/i18n/locales/en.json
@@ -898,7 +898,6 @@
     "reasoningEffort": "Reasoning",
     "reasoningOff": "Off",
     "reasoningAdaptive": "Adaptive",
-    "reasoningNone": "None",
     "reasoningMinimal": "Minimal",
     "reasoningLow": "Low",
     "reasoningMedium": "Medium",
--- a/apps/web/src/i18n/locales/ja.json
+++ b/apps/web/src/i18n/locales/ja.json
@@ -869,7 +869,6 @@
     "reasoningEffort": "推論",
     "reasoningOff": "オフ",
     "reasoningAdaptive": "アダプティブ",
-    "reasoningNone": "なし",
     "reasoningMinimal": "最小限",
     "reasoningLow": "低い",
     "reasoningMedium": "中くらい",
--- a/apps/web/src/i18n/locales/zh.json
+++ b/apps/web/src/i18n/locales/zh.json
@@ -898,7 +898,6 @@
     "reasoningEffort": "推理",
     "reasoningOff": "关闭",
     "reasoningAdaptive": "自适应",
-    "reasoningNone": "无",
     "reasoningMinimal": "极简",
     "reasoningLow": "低",
     "reasoningMedium": "中",
--- a/apps/web/src/pages/bots/components/model-options.vue
+++ b/apps/web/src/pages/bots/components/model-options.vue
@@ -460,6 +460,7 @@ const nativeAvailableEfforts = computed(() => {
   return availableEffortsForMode(
     resolveThinkingMode(activeModel.value.config),
     resolveEffortLevels(activeModel.value.config, activeClientType.value),
+    activeClientType.value,
   )
 })
 
--- a/apps/web/src/pages/bots/components/reasoning-effort.test.ts
+++ b/apps/web/src/pages/bots/components/reasoning-effort.test.ts
@@ -8,10 +8,18 @@ describe('resolveEffortLevels', () => {
     }, 'openai-completions')).toEqual([REASONING_EFFORT_DISABLE, 'low', 'medium', 'high'])
   })
 
-  it('drops none, which is a provider wire value rather than a declaration', () => {
+  it('rewrites the legacy off spelling instead of dropping it', () => {
+    // A config still advertising "none" describes a model that can be turned off.
+    // Dropping it would hide Off from a model that supports it.
     expect(resolveEffortLevels({
       reasoning_efforts: ['none', 'low', 'medium', 'high'],
-    }, 'openai-completions')).toEqual(['low', 'medium', 'high'])
+    }, 'openai-completions')).toEqual([REASONING_EFFORT_DISABLE, 'low', 'medium', 'high'])
+  })
+
+  it('collapses both spellings of off into one entry', () => {
+    expect(resolveEffortLevels({
+      reasoning_efforts: ['none', REASONING_EFFORT_DISABLE, 'low', 'high'],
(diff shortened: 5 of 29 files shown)


chen-ran (approved):

chen-ran (changes requested):
这次只改了 none 到 disable 的新约定，却没有在模型配置的读入边界把存量和上游返回的 none 统一转换成 disable。

HoneyBBQ:
## Current stack

The timeline above references several closed PRs, which makes this look deeper
than it is. The live stack is four:

```
main
└── #966  refactor/unify-reasoning-off-token   ← this PR
    └── #991  internal/reasoning leaf package + subagent fix (#983)
        └── #992  /reasoning reads the model
            └── #993  API + web + Gemini
```

#984–#988, #990 and #994 are closed. The first six were compressed into #991–#993
before anyone reviewed them, so nothing was lost; #994 was a capability upgrade
(swapping the sync source to models.dev) that turned out not to fix anything
anyone had reported, so it was dropped with its findings recorded on the PR.

## Rebased onto main

`main` had moved twelve commits ahead and this branch had gone `CONFLICTING`. The
whole stack is now rebased and all four are `MERGEABLE`. Three conflicts, all of
the same shape — main added something adjacent to what this work changed:

- `internal/models/types.go` — main extracted `ValidateCompatibilities`; kept
  alongside the vocabulary rewrite below it.
- `native.RunConfig` — main added seven `Context*` fields while this stack
  replaced five `Reasoning*` fields with one. I diffed the two field sets before
  merging to confirm no overlap, rather than eyeballing the block.
- `packages/sdk/src/index.ts` — a generated barrel file, so it was regenerated
  rather than hand-merged.

`go build ./...`, `go vet ./...`, `go test ./...` green after each step; swagger
and the TypeScript SDK regenerated.

The human-QA notes on #991–#993 were written before this rebase. I judged them
still valid because no behaviour changed — every conflict resolution preserved
both sides — but that is a judgement, not a fact. Say the word and I will restore
the `No human QA` line on all three and re-walk the paths on a dev stack.

chen-ran (approved):

PR #865, opened 2026-07-26
feat(session-runtime): implement durable distributed session runtime

chen-ran:
## What changed

- Add `docs/design/session-runtime-requirements.md` as the correctness contract for admission, ownership and fencing, reconnect observability, multi-subscriber convergence, cross-instance controls, durability, explicit turn/decision identity, and the single-instance versus distributed deployment boundary.
- Add an `integration`-tagged black-box suite that uses only public HTTP/WebSocket APIs, two real `cmd/agent` Server processes, a shared PostgreSQL database, and a deterministic OpenAI-compatible fake model.
- Add an isolated Compose overlay and mise tasks for starting, running, and stopping the acceptance environment. Startup now waits for both Server instances to become healthy; the destructive owner-crash case remains opt-in.

## Why

The existing Manager and backend tests validate internal algorithms, but they do not prove that the public WebSocket path, persistence path, and multi-instance process boundary provide the same guarantees. This change puts the required behavior in one document and makes the current integration gaps reproducible before more Session Runtime code is added.

This PR does not wire Session Runtime into the public WebSocket handler or implement the missing guarantees. The acceptance package is gated by the `integration` build tag and `MEMOH_SESSION_RUNTIME_ACCEPTANCE`, so normal test runs are unchanged.

## Main baseline

Verified against `main@b49939582`:

| Requirement | Result | Observation |
| --- | --- | --- |
| `SR-BASE-001` | Pass | One model execution; one user turn and one assistant turn persisted. |
| `SR-OBS-001` | Expected fail | `runtime_subscribe` is rejected as an unknown message type. |
| `SR-CTL-001` | Expected fail | The secondary Server returns no `control_ack` and does not cancel the owner. |
| `SR-ADM-001` | Expected fail | One invocation executes twice and persists two matching user messages. |
| `SR-OWN-001` | Expected fail | Two Servers enter the model concurrently for one Session (`max_active=2`). |
| `SR-DUR-001` | Expected fail | After `SIGKILL`, the input is absent from history and no durable runtime snapshot is available. |

`SR-OBS-003` now defines the multiple-page/multiple-subscriber convergence requirement. Its dual-connection black-box case is explicitly listed as follow-up work. Fault-injection coverage for terminal replay and decision recovery is also deferred.

## Validation

Passing checks:

- repository commit hooks: Go tests and lint
- `go test -race ./internal/agent/runtime/session ./internal/runtimefence`
[...]

PR #969, opened 2026-08-06
feat(context): run-keyed context lifecycle persistence and read ownership

ChrAlpha:
## What changes

Context lifecycle diagnostics are now keyed by admitted run identity and persisted only after the matching durable terminal fence. Native, ACP, discuss, scheduled, heartbeat, resumed-decision, and subagent paths carry the admitted run ID through execution and record one content-light terminal snapshot for `completed`, `failed_provider`, or `aborted` outcomes.

The session runtime publishes authoritative post-fence terminal observations, and its bounded repair path reconciles missed observations without letting stale owners overwrite newer outcomes. Snapshot and error-code authority are independent: authoritative application snapshots can improve recovered metadata while richer stable error codes survive stale repair.

The session lifecycle endpoint owns reads from the run-keyed table. It falls back to legacy assistant metadata only when the run query returns zero rows; query failures are returned as stable Problem Details codes rather than silently changing read sources. Snapshots contain only version, manifest view/counts, and an optional assistant-message association—never prompts, message bodies, memory payloads, tool definitions, or private diagnostics.

## Scope

This PR contains run identity plumbing, terminal lifecycle persistence and reconciliation, the run-keyed read API, stable error vocabulary and localization, PostgreSQL schema/query work, generated OpenAPI/SDK contracts, and acceptance/integration coverage. It is now standalone on upstream main after PR1 (#959) and PR5 (#960) merged.

Runtime budget-failure and fallback classification, lifecycle aggregation, latest-composition consumers, and tool-roster attribution are intentionally excluded. Those seams arrive in a later PR; the schema forward-declares `failed_budget` and `fallback`, but this PR does not classify live runs into them. No Web page, store, composable, or route consumes the endpoint in this change.

## Migration

The migration is `0130_context_lifecycles`, retaken after upstream's `0129_bot_workdirs`. The paired up/down migration is idempotent, the canonical `0001_init.up.sql` contains the same final table/index/RLS definition, all integration references use `0130`, and regenerated sqlc output is clean.

## Restack notes

The 42 original PR6 commits were replayed with `git -c commit.gpgsign=false rebase --onto upstream-main-snapshot acc97a2c59f2a1fea667a164a6cc5a1dcb893af3 land/pr6-lifecycle`. The target snapshot is `24c98175c583c7958e2c41dab7e8044ea10467bd`, which contains the merged PR1 (#959) and PR5 (#960) squashes. The immutable old PR5 tip is the critical `--onto` boundary: the local `land/pr5-runtime-decision` label had already moved during its own restack, so using that mutable label would have replayed unrelated commits.

Five focused correctness commits were added after adversarial review:

- `b205b7025 fix(agent): preserve spawned step lifecycle durability`
- `24b3b7153 fix(agent): reject aborted spawned runs`
- `7d3d9c6e7 fix(agent): detach discuss terminal persistence`
- `7067af11a fix(agent): align spawned abort outcomes`
- `3c4c0852a fix(agent): linearize spawned terminal outcomes`

The final fix makes the session runtime's serialized terminal mutation the single source of truth for spawned-run outcomes. Native attempts buffer terminal events until their disposition is known, interrupted or completed durable checkpoints suppress retry, and running-task cancellation is a nonterminal stop request until the runtime publishes the actual outcome. `HandleAgentEventWithStatus` returns the status from the exact snapshot produced by the terminal mutation, avoiding a fallible second read; a fault-injection test proves a queued post-mutation load failure remains unconsumed. Legacy `HandleAgentEvent` behavior remains intact.

The background/runtime changes are a necessary upstream-conflict adaptation, not an expansion into later lifecycle consumers: upstream #941 made running-agent kill/wait/event behavior observable, so PR6 must linearize those paths to keep the background result, session runtime, mutation ledger, and persisted lifecycle terminal in agreement. Runtime budget/fallback classification and PR8 aggregation or tool-roster ownership remain excluded.

The resulting merge-base is exactly `24c98175c583c7958e2c41dab7e8044ea10467bd`; the branch contains 47 conventional non-merge commits across 87 changed paths. The old PR5 tip is not an ancestor of the final branch, confirming PR6 is standalone. The pre-restack backup remains at `refs/backup/restack-pr6-before-main-24c98175` (`12b615e08750a1ba2b0dc1bd41fa1c57922c06f8`).

Conflict resolution preserves upstream behavior from #944 and #941: interrupted-step checkpoints, ACP prompt draining and durable projection cleanup, live subagent observation, rich subagent turn identity, and incremental step persistence. Spawned complete and interrupted steps now share the run's lifecycle holder, persist recoverable lifecycle metadata, and associate the durable assistant message ID. Internal provider/loop aborts cannot be reported as successful spawned runs, while owning cancellation causes remain authoritative. Native discuss terminal storage and publication retain the admitted fence while detaching external cancellation. Fallback terminal persistence remains available when incremental persistence is not installed. The frozen intent oracle remained clean and read-only at `context-restack@de27164f4aa37c9cd8301f96fa5b37701b44d8fe`.

No push or PR creation was performed.

## Verification

The following checks passed on final head `3c4c0852a9779ecd9e4edd12856bbe4391681f5a`:

- `MISE_TRUSTED_CONFIG_PATHS=$PWD mise run sqlc-generate`
- `MISE_TRUSTED_CONFIG_PATHS=$PWD mise run swagger-generate`
- `pnpm --config.verify-deps-before-run=false run generate-sdk`
- `pnpm --config.verify-deps-before-run=false --filter @memohai/sdk typecheck`
- `go build ./...`
- `go test -count=1 ./...`
- `golangci-lint run` over every touched Go package: application, background, context fragments, native/session runtimes, tools, errors, DB/sqlc/store, handlers, and spec
- `go test -race -count=1 ./internal/agent/application`
- `go test -race -count=1 ./internal/agent/background ./internal/agent/runtime/native ./internal/agent/runtime/session ./internal/agent/tool`
- PostgreSQL 18 integration selection: migration down/up round trip, canonical-init parity, content-light query round trip, terminal upsert convergence, and terminal-run reconciliation
- `git diff --check`

The PostgreSQL gate used disposable container `memoh-pr6-restack-pg18` (`postgres:18`) on host port 25549. Readiness and all five selected integration tests passed in 5.392 seconds. The exact container was force-removed afterward, intentionally discarding its temporary test database; the container name is absent and port 25549 is clear.

Generator and topology checks left the repository clean. The merge-base equals the target snapshot, the range has 47 commits, zero merge commits, 87 changed paths, and no non-conventional subjects. `0129_context_lifecycles` has no references, the frozen oracle is clean, the branch has no upstream, and local `main` remains at `117bb3762abc0923aee65877d616105c13e22cc2` with no restack-era reflog movement.

The aggregate `mise run sdk-generate` wrapper reached its pnpm dependency step, where pnpm 11 stopped on `ERR_PNPM_IGNORED_BUILDS` for `protobufjs@7.6.5`. No build script was approved and the temporary pnpm workspace placeholder was removed. The repository's installed OpenAPI generator and SDK TypeScript compiler were then run directly as listed above and passed.

## Landing stack

| PR | Scope |
| --- | --- |
| PR1 | Fragments-first authoritative compiler with byte-equivalent provider output |
| PR2 | System-section retention, drop policy, grouped rendering, and capability gating |
| PR3 | Unified `ContextBudgetPlan` and budget-aware behavior for all five modes |
| PR4 | Hook authority split and provider-attempt preflight |
| PR5 | Runtime decision correctness and fencing for #865 |
| PR6 | Run-keyed lifecycle persistence and read ownership |
| PR7 | Prompt shrink, baselines, and repository-level gates |
| PR8 | Cross-stack lifecycle, budget, fallback, and tool-roster integration seams |


chen-ran:
Could you clarify why this needs a separate context_lifecycles table instead of extending session_runs or deriving the data from session_runs and message metadata? What follow-up features will consume this table, and are they planned to land soon?

ChrAlpha:
`session_runs` and `context_lifecycles` answer different questions under different write disciplines, and merging them breaks invariants on both sides:

- **Coverage** — not every run that needs a lifecycle row is an admitted ledger run. Direct runs execute with no durable `session_runs` row (the unfenced path in `service_run_lifecycle.go`), and the table also records subagent, discuss, scheduled, and heartbeat outcomes. Extending `session_runs` leaves those runs nowhere to land.
- **Ledger invariants** — `session_runs` is the admission/fencing substrate; its header comment deliberately restricts it to admission, ownership/fencing, decisions, and terminal transitions, with no snapshot columns. A row takes several fenced updates per run, so attaching the JSONB manifest snapshot would copy it on every MVCC rewrite. The lifecycle repair path also has its own authority split (an authoritative snapshot can improve a recovered row while a richer stable error code survives stale repair), which doesn't map onto the ledger's fenced single-writer discipline.
- **Vocabulary** — `session_runs.state` is run coordination (`accepted … lost`); lifecycle `status` classifies context outcomes: `fallback` is a run that *completed* with degraded context, and `failed_budget`/`failed_provider` split the ledger's single `failed`. The two aren't 1:1, so it can't be another CHECK on the same column.

Deriving from `session_runs` + message metadata is exactly what the legacy fallback in this PR does (`metadata ? 'context_lifecycle'` over assistant messages), and it's insufficient by construction: `failed_provider`/`aborted` runs aren't guaranteed to persist an assistant message, so derivation goes blind precisely for the outcomes this PR exists to make observable, and `session_runs` carries error codes but no manifest snapshot. The fallback is kept only for pre-table rows, gated to zero-row results, with query failures surfaced as stable Problem Details so the read source never flaps silently.

Consumers: the run-keyed session lifecycle endpoint in this PR is the read owner today. PR8 (last row of the landing-stack table) adds what motivated the forward-declared statuses: runtime budget-failure/fallback classification writing `failed_budget`/`fallback`, lifecycle aggregation, latest-composition consumers, and the web inspector wiring. PR3/PR4/PR7 are queued ahead of it and the stack is landing in order, so near-term. Declaring the two statuses now avoids re-migrating the CHECK constraint in PR8.

chen-ran (approved):

Issue #983, opened 2026-08-10
[Bug]: subagent spawn drops the entire reasoning configuration


HoneyBBQ:
## Problem

Subagents run with no reasoning configuration at all. Not a downgraded one — none.
Whatever the parent bot resolved (adaptive thinking, a chosen effort tier, or an
explicit "off") is dropped at the spawn boundary, and the subagent falls back to
whatever the provider does by default.

The decision has **two carriers**, and spawn breaks both:

**Carrier A — provider construction.** `NewSDKChatModel` welds Anthropic's
`thinking{type:"adaptive"}` or `thinking{type:"enabled", budget_tokens:N}` onto the
provider instance (`internal/models/sdk.go:121-132`). The parent passes
`ReasoningConfig` when building it (`internal/agent/application/service.go:731`).
The subagent's `resolveModel` builds its own `SDKModelConfig`
(`internal/agent/tool/subagent.go:1624-1631`) and simply omits that field.

**Carrier B — per-request options.** The parent flattens the config into five
`native.RunConfig` fields (`service.go:757-761`) which `native/agent.go:935-945`
reassembles before calling `BuildReasoningOptions`. On the spawn path,
`tools.SpawnRunConfig.ReasoningEffort` (`internal/agent/tool/subagent.go:50`) is
**declared and never assigned anywhere in the production tree** — so
`spawn_adapter.go:131` faithfully carries an empty string, and the other four fields
have no field to travel in to begin with.

There is no recomputation to fall back on: `native/agent.go:935-945` is a mechanical
reassembly that consults neither the model nor settings.

## Why losing the flags is total, not partial

`openAIEffortOptions` (`internal/models/sdk.go:233-256`) switches on `Active` and
`Disabled` only, with `default: return nil`. With both false, nothing is sent even if
an effort string somehow survived. `Effort` is a dependent field — losing the
switches disables it too.

## Effect, parent set to Claude 4.6 + high + adaptive

| Case | Parent sends | Subagent sends | Result |
|---|---|---|---|
| Anthropic adaptive (4.6+) | `thinking{adaptive}` + `output_config.effort:"high"` | neither | thinking silently off |
| Anthropic legacy (≤4.5) | `thinking{enabled, budget_tokens:50000}` | no thinking block | extended thinking off |
| OpenAI, can disable | `reasoning_effort:"none"` when off | nothing → provider default (**on**) | user turned reasoning off and still pays for it |
[...]

Issue #982, opened 2026-08-10
refactor(reasoning): single source of truth for model reasoning capability


HoneyBBQ:
## Problem

"Can this model turn thinking off, and which tiers does it offer?" is answered
independently in three places, and the only layer that actually knows — the provider
adaptor — is asked by none of them.

- **Web picker**: `apps/web/src/pages/bots/components/reasoning-effort.ts` derives it
  in TypeScript from `(thinking_mode, reasoning_efforts, client_type)`.
- **`/reasoning` command**: `internal/command/reasoning.go:15` hardcodes
  `off, low, medium, high, xhigh` and never looks at the model at all.
- **Backend resolver**: seven pure functions in
  `internal/agent/application/service.go` (`resolveReasoningConfig`, `pickEffort`,
  `offEffortFor`, `effectiveReasoningEfforts`, `anthropicEffortEra`, …).

The consequences are not hypothetical. #966 spent four commits on this area and three
of them close a hole the previous one opened: unifying the off token made `disable`
advertisable, which let `pickEffort` hand it back as an *active* tier; normalizing
that required rewriting the legacy spelling at both `ModelConfig` boundaries in Go
*and* TypeScript; and gating the picker on the advertised token hid Off from every
Claude model, which needed a hardcoded `IMPLICIT_OFF_CLIENT_TYPES` client-type list
to bring back.

Two surveys of the repo turned up what that pattern has already cost:

- **DeepSeek V4 had no Off in the web picker** while `/reasoning off` worked fine —
  the compat layer supports turning it off, the catalog just has no way to say so.
  (Fixed in #966; MiniMax deliberately left alone, since models.dev and OpenRouter
  both report M2.x as always-on and adding the token there would legitimize a
  control the model cannot honor.)
- **`chat-pane.vue:1753` omits the third argument** to `availableEffortsForMode`, so
  the chat composer and the bot settings page disagree about whether Claude can be
  turned off.
- **Four copies of the effort vocabulary** (`models.validReasoningEfforts`,
  `capabilities.effortOrder`, `command.reasoningChoices`, the frontend's
  `KNOWN_EFFORTS`), plus a fifth hardcoded in an i18n usage string that still lists a
  value no longer selectable.
- **Two implementations of max→xhigh** (`normalizesMaxReasoningEffort` and
  `openAIWireEffort`), whose own comment calls the duplication defence-in-depth.
- **Seven "keep in sync" comments** spanning Go and TypeScript, one of which points
[...]

PR #984, opened 2026-08-10
refactor(reasoning): give the capability layer one home

HoneyBBQ:
Stacked on #966. First step of #982.

## What this does

Moves every reasoning capability decision into `internal/reasoning`, a leaf
package that imports nothing else in the tree. Behaviour is unchanged — this is a
move, not a rewrite.

The property worth reviewing for is that `OptionsFor` (what a picker may offer)
and `ResolveConfig` (what a call sends) share their internals. They were separate
answers before, in separate languages, and every bug in this area was a
disagreement between them. `TestOptionsDefaultEffortMatchesWhatResolveWouldSend`
asserts they agree across five catalogs.

Off-ness also stops being a sentinel inside the tier list. It is
`Options.CanDisable` now, so no consumer has to remember to filter `disable` out
of a list of efforts — the omission that let an *active* config resolve to "off"
twice in #966.

## The evidence that behaviour did not change

The resolver's table-driven tests in `internal/agent/application` pass with
**their assertions untouched**; only symbol names moved. `TestOffEffortFor` moved
to the leaf package verbatim.

## Also removed

| Before | After |
|---|---|
| 4 copies of the effort vocabulary | one, in `internal/reasoning` |
| `native.ModelConfig`, `native.ReasoningConfig` | deleted — both unreferenced |
| `settings.isValidReasoningEffort` | `hasReasoningEffortValue` — it only checked for a non-blank value, while the identically named function in `models` enforced the real vocabulary |
| `RunConfig`'s five flat reasoning fields | one `*models.ReasoningConfig` |

`internal/models` keeps forwarding aliases (`models.ReasoningEffortDisable` etc.)
so existing call sites migrate on their own schedule rather than in one sweep.

## Why the flat fields mattered

They were split for a serialization constraint that does not exist —
`sdk.GenerateParams` is never marshalled — and splitting them is how the subagent
spawn path came to carry one field and silently drop four (#983). Carrying the
config whole makes that class of loss unrepresentable. The spawn path still needs
its own resolution against the subagent's own model; that lands next, now that
`internal/agent/tool` can import a leaf package.

## Verification

`go build ./...`, `go vet ./...`, `go test ./...` all green; `gofmt` clean.

### DIFF
--- a/internal/agent/application/service.go
+++ b/internal/agent/application/service.go
@@ -43,6 +43,7 @@ import (
 	"github.com/memohai/memoh/internal/models"
 	"github.com/memohai/memoh/internal/oauthctx"
 	"github.com/memohai/memoh/internal/providers"
+	"github.com/memohai/memoh/internal/reasoning"
 	"github.com/memohai/memoh/internal/settings"
 	"github.com/memohai/memoh/internal/workspace"
 )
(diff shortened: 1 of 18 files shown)


HoneyBBQ:
Superseded by #991, which carries the same commits plus the review fixes. Closing to keep one review surface per change rather than six — none of these had been reviewed, so nothing is lost.

PR #985, opened 2026-08-10
fix(agent): resolve reasoning for subagents instead of dropping it

### LINKED ISSUE #983: [Bug]: subagent spawn drops the entire reasoning configuration

HoneyBBQ:
Fixes #983. Stacked on #984.

## The bug

Subagents ran with **no** thinking configuration — not a downgraded one, none.
Both carriers of the decision were empty:

| Carrier | Where it broke |
|---|---|
| Provider construction (Anthropic's `thinking{...}` block) | `resolveModel` built `SDKModelConfig` without `ReasoningConfig` |
| Per-request options (`reasoning_effort`) | `SpawnRunConfig.ReasoningEffort` was declared but **never assigned anywhere in the production tree** |

`BuildReasoningOptions` switches on `Active`/`Disabled`, so a config with both
false emits nothing. The parent's choice — on *or* off — silently became
"whatever the provider does by default".

It erred in both directions: a parent on Claude 4.6 at high effort spawned a
subagent that did not think, and a user who explicitly turned reasoning **off**
still paid for it on every OpenAI-format model that can be turned off.

## Why resolve instead of inherit

`spawn_agent` accepts `model_id`/`provider`, so a subagent may run a model whose
advertised tiers and off-ability differ from the parent's. Copying the parent's
resolved config would send a tier the subagent's model does not offer, or claim
an off switch it does not have.

`resolveModel` now resolves against the model it just resolved, using the bot's
stored effort as the on/off source. This is only possible after #984 moved the
resolver into a leaf package — `internal/agent/tool` cannot import
`internal/agent/application`.

A missing settings service still yields nil, which is the same "no decision" as
before. That is a floor, not a target: reasoning falls back to the provider
default rather than failing the run.

## Tests

The characterization tests written while diagnosing this are kept, with
assertions flipped from "the decision is dropped" to "the decision arrives" —
including the control test showing the parent turn always sent it, which is what
localized the defect to the spawn boundary.

One test is new and worth reviewing:
`TestResolveSubagentReasoningFollowsTheSubagentModel` resolves the same stored
effort against two catalogs — one that can be turned off, one that cannot — and
asserts the answers differ. That is precisely the case naive inheritance gets
wrong.

### DIFF
--- a/internal/agent/runtime/native/spawn_adapter_reasoning_test.go
+++ b/internal/agent/runtime/native/spawn_adapter_reasoning_test.go
@@ -1,0 +1,226 @@
+package native
+
+import (
+	"context"
+	"testing"
+
+	sdk "github.com/memohai/twilight-ai/sdk"
+
+	tools "github.com/memohai/memoh/internal/agent/tool"
+	"github.com/memohai/memoh/internal/models"
+)
+
+// The reasoning decision on the subagent spawn path (#983).
+//
+// A subagent used to run with no thinking configuration at all: SpawnRunConfig
+// carried a lone ReasoningEffort string that production code never assigned, and
+// the four flags that actually drive the wire had no field to travel in. Since
+// BuildReasoningOptions switches on Active and Disabled, a config with both false
+// emits nothing — so the parent's choice, on or off, silently became "whatever
+// the provider does by default".
+//
+// SpawnRunConfig now carries the whole *models.ReasoningConfig, resolved against
+// the subagent's own model. These tests assert it reaches the wire.
+
+// spawnReasoningProvider records the GenerateParams the SDK finally sends, so a
+// test can assert on the wire-level reasoning shape.
+type spawnReasoningProvider struct {
+	name   string
+	params sdk.GenerateParams
+	calls  int
+}
+
+func (p *spawnReasoningProvider) Name() string {
+	if p.name == "" {
+		return "openai-completions"
+	}
+	return p.name
+}
+
+func (*spawnReasoningProvider) ListModels(context.Context) ([]sdk.Model, error) {
+	return nil, nil
+}
+
+func (*spawnReasoningProvider) Test(context.Context) *sdk.ProviderTestResult {
+	return &sdk.ProviderTestResult{Status: sdk.ProviderStatusOK}
+}
+
+func (*spawnReasoningProvider) TestModel(context.Context, string) (*sdk.ModelTestResult, error) {
+	return &sdk.ModelTestResult{Supported: true}, nil
+}
+
+func (p *spawnReasoningProvider) DoGenerate(_ context.Context, params sdk.GenerateParams) (*sdk.GenerateResult, error) {
+	p.calls++
+	p.params = params
+	return &sdk.GenerateResult{
+		Text:         "ok",
+		FinishReason: sdk.FinishReasonStop,
+	}, nil
+}
+
+func (*spawnReasoningProvider) DoStream(context.Context, sdk.GenerateParams) (*sdk.StreamResult, error) {
+	return nil, nil
+}
+
+// runSpawnReasoning drives the real spawn conversion plus buildGenerateOptions
+// and returns the reasoning effort the provider actually received.
+func runSpawnReasoning(t *testing.T, providerName string, cfg tools.SpawnRunConfig) *string {
+	t.Helper()
+
+	provider := &spawnReasoningProvider{name: providerName}
+	cfg.Model = &sdk.Model{
+		ID:       "test-model",
+		Provider: provider,
+		Type:     sdk.ModelTypeChat,
+	}
+
+	rc := runConfigFromSpawnRunConfig(cfg)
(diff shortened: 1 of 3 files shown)


HoneyBBQ:
Superseded by #991, which carries the same commits plus the review fixes. Closing to keep one review surface per change rather than six — none of these had been reviewed, so nothing is lost.

PR #986, opened 2026-08-10
refactor(command): let /reasoning read the model instead of a fixed list

HoneyBBQ:
Stacked on #985. Part of #982.

## The change

`/reasoning` offered `off, low, medium, high, xhigh` on every model, having never
consulted one. Wrong in both directions:

- **Off appeared on models that cannot be turned off**, where picking it does
  nothing the provider honours
- **`minimal` and `max` never appeared** on the models that advertise them

It was also the fourth copy of the effort vocabulary in the tree, and the one
furthest from the resolver that decides what actually gets sent.

The command now asks the bot's current chat model through the same `OptionsFor`
the picker and the resolver share (#984). A typed level is validated the same
way, so a tier the model does not offer is refused at the command rather than
stored and quietly ignored at call time.

## Fallback

A model that cannot be resolved — no chat model set, or the lookup fails — falls
back to the common `off/low/medium/high` ladder rather than rendering an empty
picker. Breaking the command because a model row is missing is a worse answer
than offering the tiers every reasoning model has.

## Also

The usage strings in all three locales still advertised `none`, which stopped
being selectable when #966 gave off a single name.

## Tests

`TestReasoningChoicesFollowTheModel` covers the three cases that used to be one:
a model that can be turned off, one that cannot, and an unresolvable one.
`TestAcceptsEffortRejectsTiersTheModelDoesNotOffer` covers the typed-level half.

### DIFF
--- a/internal/command/reasoning.go
+++ b/internal/command/reasoning.go
@@ -2,32 +2,41 @@ package command
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/memohai/memoh/internal/i18n"
 	"github.com/memohai/memoh/internal/models"
+	"github.com/memohai/memoh/internal/reasoning"
 	"github.com/memohai/memoh/internal/settings"
 )
 
-// reasoningChoices are the selectable reasoning levels. "off" disables thinking;
-// the rest enable it at that effort. Only "off" represents the disabled state —
-// listing "none" here too offered the same state under a second name.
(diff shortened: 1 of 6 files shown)


HoneyBBQ:
Superseded by #992, which carries the same commits plus the review fixes. Closing to keep one review surface per change rather than six — none of these had been reviewed, so nothing is lost.

PR #987, opened 2026-08-10
feat(api): ship resolved reasoning options with every chat model

HoneyBBQ:
Stacked on #986. Part of #982.

## What ships

Every chat model in the API now carries the resolved answer instead of the raw
ingredients:

```json
"reasoning": {
  "supported": true,
  "can_disable": true,
  "efforts": ["low", "medium", "high"],
  "default_effort": "medium"
}
```

It comes from the same `OptionsFor` the resolver and `/reasoning` already read
(#984), so what a client renders and what the wire receives are one computation.

## Why the frontend could not get this right

The web picker derived capability from `thinking_mode` + `reasoning_efforts` +
`client_type`, in TypeScript, from a projection that **did not include everything
the answer depends on** — it cannot see `chat_completions_compat` at all, which
is why DeepSeek's Off went missing. It also joined providers itself in three
separate places to learn a client type, and [one of them forgot the
argument](https://github.com/memohai/Memoh/blob/main/apps/web/src/pages/home/components/chat-pane.vue#L1753),
so the chat composer and the settings page disagreed about Claude.

## Design notes

- **The shape follows ACP's existing `ReasoningState`** (`available_efforts` /
  `current_effort` / `supported`) rather than inventing a second pattern for
  native models.
- **Computed in the handler, not stored** — it depends on the provider's client
  type, one join away from the model row, and it is a projection of the catalog
  rather than a fact about it.
- **A provider that cannot be read** still yields the model's own tiers with the
  wire policy missing. That beats omitting the field and sending clients back to
  deriving it.
- **A model with no thinking concept sends `{"supported": false}`** rather than
  nothing, so a client can tell "cannot think" from "the server did not say".

## Verification

`mise run swagger-generate` run; `reasoning.Options` is in the spec with
snake_case fields.

The TypeScript SDK regeneration lands with the frontend change that consumes it.

### DIFF
--- a/internal/handlers/models.go
+++ b/internal/handlers/models.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"context"
 	"errors"
 	"log/slog"
 	"net/http"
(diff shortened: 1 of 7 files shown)


HoneyBBQ:

PR #988, opened 2026-08-10
refactor(web): render reasoning options instead of deriving them

HoneyBBQ:
M4 of #982: the frontend stops deriving reasoning capability and renders what the API resolved in #982's M3 (`a980d84`, this PR's base).

## What went away

`apps/web/src/pages/bots/components/reasoning-effort.ts` shrinks from 162 lines to 64. Deleted:

- `KNOWN_EFFORTS`, `DECLARABLE_EFFORTS`, `LEGACY_OFF_EFFORT` — the frontend's copy of the effort vocabulary
- `MAX_NORMALIZED_CLIENT_TYPES`, `IMPLICIT_OFF_CLIENT_TYPES` — the two client-type special cases
- `resolveThinkingMode`, `resolveEffortLevels`, `nearestEffortToMedium`, `availableEffortsForMode`, `canTurnOff`
- all five "keep in sync" comments, one of which pointed at `internal/conversation/flow/resolver.go` — a file that no longer exists

Kept, because they are rendering rather than derivation: `REASONING_EFFORT_DISABLE`, `EFFORT_LABELS`, `EFFORT_OPACITY`.

Two helpers replace the ten deleted symbols, and neither decides anything:

- `selectableEfforts(options)` — the tiers, with Off hoisted to the front when `can_disable`
- `reconcileStoredEffort(stored, options)` — a stored value the model still offers survives, Off survives when `can_disable`, anything else lands on `default_effort`. Mirrors `reasoning.ReconcileStored`, which owns the same policy for `/reasoning`.

## Consumers

The three `activeClientType` computeds are gone along with the providers join they existed for — `model-options.vue`, `chat-pane.vue`, and `settings-interaction-card.vue` all read `model.reasoning` now. The two stranded-effort migrations (bot settings and chat composer) collapse onto the one rule above.

**`chat-pane.vue:1753` is fixed by construction.** That call omitted `availableEffortsForMode`'s third argument (`clientType`), so the chat composer and the bot settings page disagreed about whether Claude could be turned off. There is now one server-side answer and no second caller to pass the wrong arguments to.

## json tags

The brief for this PR asked me to add snake_case json tags to `reasoning.Options` — the base commit already did (`supported` / `can_disable` / `efforts,omitempty` / `default_effort,omitempty`), so no change was needed. This PR regenerates the SDK against that spec, which the base had not done: `ModelsGetResponse` now carries `reasoning?: ReasoningOptions`.

## Dead i18n keys

Removed from `en`/`zh`/`ja` after grepping for consumers: `bots.settings.reasoningEnabled`, `bots.settings.reasoningEffort`, `bots.settings.reasoningEffort{None,Low,Medium,High,XHigh}`, `bots.overview.reasoningBadge`. Note `chat.reasoningEffort` is a *different* key, still used by the picker's aria-label, and stays.

## Verification

- `vitest --project web`: 109 files, 1037 tests green. The `resolveEffortLevels` / `availableEffortsForMode` / `nearestEffortToMedium` suites are deleted — they tested logic that is no longer in TypeScript — and replaced with coverage of the two rendering helpers.
- `mise run lint` clean (ESLint, golangci-lint, UI contract).
- Go: `internal/handlers`, `internal/models`, `internal/reasoning` green, including the base's new `models_reasoning_test.go`.
- Grepped: no frontend code reads `client_type` or `thinking_mode` for a reasoning decision. Remaining hits are provider-configuration UI and the model-config editor, which *writes* `thinking_mode` rather than deriving capability from it.

The 5 failing `apps/desktop` tests (`server-connection`, `remote-runtime`) fail identically on the base commit and are unrelated.


HoneyBBQ:

PR #990, opened 2026-08-11
feat(models): make Gemini's thinking control real

HoneyBBQ:
Stacked on #988. Closes the Gemini half of #982 / memohai/twilight-ai#41.

## The control was never connected

The picker has offered Gemini three tiers since the day it was added, and none of
them ever left the building: the Google provider had no thinking field on its
request, so `BuildReasoningOptions` returned nil and the selection was a placebo.
memohai/twilight-ai#42 added the wire; this connects it.

Requires the dependency bump included here (`twilight-ai@70292145`).

## Why the catalog has to declare the dialect

Gemini 2.5 takes `thinkingConfig.thinkingBudget`; 3.x takes
`thinkingConfig.thinkingLevel`; **sending both is a 400**. The tiers look
identical from the outside, so the wire shape is not inferable from what a model
advertises — and the alternative to declaring it is sniffing the model id, which
is the practice this whole effort exists to remove.

So the catalog says it: `reasoning_dialect` plus the budget bounds, per model.

## Proportional budgets, not a fixed table

A tier maps across each model's **own** range:

| model | range | `high` |
|---|---|---|
| 2.5 Pro | 128–32768 | 26240 |
| 2.5 Flash | 0–24576 | 19660 |
| 2.5 Flash-Lite | 512–24576 | 19763 |

A fixed table cannot be right for models with different ceilings, and the vendors
publish no tier-to-token mapping at all — they describe tiers as *relative
allowances rather than guarantees*. Cline and Cherry Studio independently landed
on the same proportional approach.

## Off-ability falls out of the bounds

A zero budget is legal exactly where the floor is zero:

- **2.5 Flash** (floor 0) can be turned off, and now advertises it
- **2.5 Pro** (floor 128) cannot; asking it to would earn a 400
- **3.x** has no off value at all — `minimal` is the floor — so a disabled model
  sends nothing and lets the provider default stand

## `includeThoughts`

Set whenever thinking is on. Without it the API runs and **bills** the thinking
but emits no thought parts, so the reasoning stream stays empty for a user who
can see they are paying for it.

## Guards

This design trades sniffing for a declaration, and declarations rot. Two template
tests:

1. A Google model declaring a thinking mode without a dialect fails.
2. A budget model whose advertised off-ability disagrees with its floor fails.

Both were confirmed to fail against a deliberately broken catalog before being
kept — a guard that has never failed is not known to work.

## synccaps

It was stripping the `minimal` tier: LiteLLM reports `supports_minimal` for one
model family while Gemini 3.x takes it as the floor. Same shape of gap as the
`disable` token, same treatment — registry silence is not evidence of absence.
`go run ./cmd/synccaps -check` is clean.

## Verification

`go build ./...`, `go vet ./...`, `go test ./...` green; `gofmt` clean; swagger
regenerated.

### DIFF
--- a/cmd/synccaps/main.go
+++ b/cmd/synccaps/main.go
@@ -185,14 +185,20 @@ func applyToModel(model *yaml.Node, mode string, efforts []string) bool {
 	if len(wantEfforts) == 0 {
 		wantEfforts = []string{"low", "medium", "high"}
 	}
-	// A hand-maintained disable token is preserved, not overwritten. LiteLLM's
-	// only off signal is the OpenAI-wire supports_none_reasoning_effort flag, so
-	// for providers whose off travels another way (DeepSeek/MiniMax toggle via
-	// chat_completions_compat) the token can only ever come from the template.
-	// Registry silence must not strip what the registry cannot know about.
-	if existing := mapValue(cfg, "reasoning_efforts"); seqContains(existing, "disable") &&
-		!slices.Contains(wantEfforts, "disable") {
-		wantEfforts = append([]string{"disable"}, wantEfforts...)
+	// Hand-maintained tokens the registry cannot know about are preserved rather
+	// than overwritten. LiteLLM's only off signal is the OpenAI-wire
+	// supports_none_reasoning_effort flag, so where off travels another way
+	// (DeepSeek/MiniMax toggle via chat_completions_compat, Gemini 2.5 Flash via a
+	// zero budget) the token can only come from the template. The same holds for
+	// minimal, which the registry reports for exactly one model family while
+	// Gemini 3.x accepts it as its floor. Registry silence is not evidence of
+	// absence.
+	if existing := mapValue(cfg, "reasoning_efforts"); existing != nil {
+		for _, token := range []string{"disable", "minimal"} {
+			if seqContains(existing, token) && !slices.Contains(wantEfforts, token) {
+				wantEfforts = append([]string{token}, wantEfforts...)
+			}
+		}
 	}
 
 	changed := false
--- a/cmd/synccaps/main.go
+++ b/cmd/synccaps/main.go
@@ -194,10 +194,16 @@ func applyToModel(model *yaml.Node, mode string, efforts []string) bool {
 	// Gemini 3.x accepts it as its floor. Registry silence is not evidence of
 	// absence.
 	if existing := mapValue(cfg, "reasoning_efforts"); existing != nil {
+		// Collected in canonical order and prepended once, so the list stays
+		// weakest-to-strongest instead of reversing with each insertion.
+		var carried []string
 		for _, token := range []string{"disable", "minimal"} {
 			if seqContains(existing, token) && !slices.Contains(wantEfforts, token) {
-				wantEfforts = append([]string{token}, wantEfforts...)
+				carried = append(carried, token)
 			}
+		}
+		if len(carried) > 0 {
+			wantEfforts = append(carried, wantEfforts...)
 		}
 	}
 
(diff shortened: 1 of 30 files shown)


HoneyBBQ:
Pushed `ff8b8b3a`, which addresses all four 🔴 and the four 🟡 from review. Three
of the 🔴 turned out to share a cause, and one of them changed my mind about the
design.

## 🔴1 — dialect lost on the import path

Correct, and the most serious of the four. `RemoteModel` had no fields for the
dialect or bounds, neither template extractor read them, and the import handler
could not pass what it never received. A Gemini model created the normal way
stored an empty dialect → tier fallback → `thinkingLevel` to a 2.5 endpoint →
400 on every request. Worse than the placebo, exactly as you said.

The registry guard missed it because it inspects YAML, and YAML was right. The
assembly is now a named function (`modelConfigFromRemote`) so a test can exercise
the path a model actually takes; I verified it fails when any single declaration
is dropped.

## 🔴2 — managed catalogs

Both halves confirmed. Copilot/Codex filtered upstream efforts through
`IsValidReasoningEffort`, where `none` is no longer valid, so upstream's own
spelling of off was discarded — while the LiteLLM path has mapped it to `disable`
all along. They normalize now. And `mergeManagedDiscoveredConfig` replaced the
list wholesale, so a hand-maintained `disable` survived until the next sync and
then vanished; `preserveDeclaredCapabilities` carries tokens discovery cannot know
about across a refresh.

## 🔴3 — right instinct, wrong axis (and mine was wrong too)

I checked the official per-model table. **"4.6+" is not a generation for this
purpose** — the split is per model, and it cuts across `adaptive`:

| | omitting `thinking` | explicit `disabled` |
|---|---|---|
| Opus 4.6 / Sonnet 4.6 / 4.7 / 4.8 | **off** | accepted |
| Sonnet 5 | **still thinks** | accepted |
| Opus 5 | **still thinks** | only at effort ≤ high; xhigh/max → 400 |
| Fable 5 / Mythos 5 | **still thinks** | **400** |

So you were right that 4.6 lost a working control — but removing the
`mode == ModeToggle` guard would have handed Fable 5 a dead switch and Opus 5 a
400. Models on either side of those lines share a mode *and* an identical tier
list, so nothing derivable separates them: `reasoning_off_support` declares it,
and `Options.EffortsWithoutOff` names the tiers that conflict.

**This also promoted a convention into a hazard.** Our adaptor expressed off by
*omitting* the field. That reads as off only through 4.8; from Opus 5 on, thinking
keeps running — billed as output tokens, counted against `max_tokens` — while the
user believes it is off. It now sends an explicit `disabled` where the model
accepts one. We ship no always-on Claude models, so this is not live; it is the
trap set for whoever adds Opus 5.

Source: `platform.claude.com/docs/en/build-with-claude/thinking-troubleshooting`
(per-model table), plus the thinking and effort pages.

## 🔴4 — MiniMax stays without the token, and the comment is what's wrong

models.dev reports `reasoning_options: []` and OpenRouter reports
`mandatory: true` — both independent sources say M2.x cannot be turned off. The
synccaps comment claiming DeepSeek *and* MiniMax both disable through compat
predates that check and is the thing in error. I left the catalog alone rather
than legitimizing a control the model may not honour; a probe against the live API
should settle it before anyone edits either.

## 🟡 all four

- `unknownLevel` no longer recites a fixed ladder (it listed `none`, and levels
  are per-model now) — all three locales
- `Update` responses carry resolved options, like the read paths
- synccaps prepends preserved tokens once instead of reversing them
- a model advertising only the off token reports no default tier rather than
  naming one it does not offer

## Verification

`go build ./...`, `go vet ./...`, `go test ./...`, `gofmt` clean,
`synccaps -check` clean, swagger regenerated. Both new guards were confirmed to
fail against deliberately broken input before being kept.

HoneyBBQ:

PR #993, opened 2026-08-11
feat(reasoning): ship resolved options to clients, and make Gemini's control real

HoneyBBQ:
Third of three. Stacked on #992. Closes the code side of #982 and the Gemini half
of memohai/twilight-ai#41.

Replaces #987, #988 and #990, plus the review fixes those attracted. They land
together because they are **one contract seen from three sides**: the server
resolves, the client renders, and Gemini is the first provider whose wire shape the
contract has to describe. Reviewing the swagger diff twice was the main cost of
keeping them apart.

Requires the dependency bump included here (`twilight-ai@70292145`).

## The API ships the answer, not the ingredients

```json
"reasoning": { "supported": true, "can_disable": true,
               "efforts": ["low","medium","high"], "default_effort": "medium" }
```

Computed in the handler, not stored: it depends on the provider's client type (one
join away from the model row) and it is a projection of the catalog rather than a
fact about it. A model with no thinking concept sends `{"supported": false}` rather
than nothing, so a client can tell "cannot think" from "the server did not say".

## The frontend could never have got this right

It derived capability from a projection that **did not include everything the
answer depends on** — it cannot see `chat_completions_compat` at all, which is why
DeepSeek's Off went missing. It also joined providers in three places to learn a
client type, and one of them omitted the argument, so the chat composer and the bot
settings page disagreed about Claude.

`reasoning-effort.ts`: 162 → 64 lines. Ten derivation symbols and all five "keep in
sync" comments gone (one pointed at a deleted file). The argument-omission bug
disappears with the code rather than needing its own fix, along with 8 dead i18n
keys.

## Gemini's control was never connected

The picker has offered three tiers since the day it was added; none ever left the
building. Which field to send is **not inferable**: 2.5 takes `thinkingBudget`, 3.x
takes `thinkingLevel`, sending both is a 400, and the tiers look identical from
outside. So the catalog declares it — the alternative is sniffing a model id.

Budgets map proportionally across each model's **own** range:

| model | range | `high` |
|---|---|---|
| 2.5 Pro | 128–32768 | 26240 |
| 2.5 Flash | 0–24576 | 19660 |

The bounds also settle off-ability, otherwise a coin flip: a zero budget is legal
exactly where the floor is zero. Flash can be turned off; Pro cannot. On the tier
dialect there is no off value at all — `minimal` is the floor.

`includeThoughts` is set whenever thinking is on: without it the API **bills** the
thinking and emits no thought parts.

## Anthropic's off-ability is not generational either

Review asked whether gating on the toggle mode had cost adaptive Claude a working
control. The official per-model table says yes — **and that dropping the gate is
not the fix**:

| | omitting `thinking` | explicit `disabled` |
|---|---|---|
| Opus 4.6 / Sonnet 4.6 / 4.7 / 4.8 | **off** | accepted |
| Sonnet 5 | **still thinks** | accepted |
| Opus 5 | **still thinks** | only at effort ≤ high; xhigh/max → 400 |
| Fable 5 / Mythos 5 | **still thinks** | **400** |

Models on either side share a thinking mode *and* an identical tier list, so
`reasoning_off_support` declares it and `EffortsWithoutOff` names the conflicting
tiers.

**This turns the omission convention into a hazard.** Expressing off by omitting the
field reads as off only through 4.8; from Opus 5 on, thinking keeps running — billed
as output tokens, counted against `max_tokens` — while the user believes it is off.
The adaptor sends an explicit `disabled` where the model accepts one. No always-on
Claude model ships today; this is the trap set for whoever adds one.

## The declarations had to survive the import path — they did not

`RemoteModel` had no fields for them, neither extractor read them, and the import
handler could not pass what it never received. A Gemini model created the normal way
stored an empty dialect → tier fallback → **400 on every request**. Worse than the
placebo.

The registry guard missed it because it inspects YAML, and YAML was right. The
assembly is a named function now so a test exercises the path a model actually
takes.

Two managed-catalog defects from the same family: Copilot/Codex discarded upstream's
own spelling of off (`none`, invalid in the new vocabulary) while the LiteLLM path
has mapped it all along; and the managed merge replaced the effort list wholesale,
so a hand-maintained `disable` survived until the next sync and then vanished.

MiniMax stays without the token deliberately: models.dev reports
`reasoning_options: []` and OpenRouter reports `mandatory: true`. The synccaps
comment claiming otherwise predates that check and is the thing in error.

## Guards, verified to fail

Both new template tests were run against a deliberately broken catalog before being
kept — a guard that has never failed is not known to work.

## Verification

`go build ./...`, `go vet ./...`, `go test ./...` green; `gofmt` clean;
`synccaps -check` clean; swagger regenerated; web suite 137 tests across the
affected pages.

---

## Human QA

Walked on a local dev stack against the live Gemini API (`gemini-3.6-flash`,
tier dialect, effort `high`).

**The picker follows the model.** Switching between models changes the offered
tiers, and a stored tier the new model does not advertise falls back to a valid
one rather than sticking or blanking. Verified across the cases that used to be
one: `claude-opus-4-6` (Off available), `claude-fable-5` (no Off, same vendor and
thinking mode), `deepseek-v4-pro` (Off available — the control that was missing
from the web picker), `gemini-3.6-flash` (no Off, since `minimal` is the floor on
the tier wire), and `minimax-m2.5` (supported, no controls at all).

**Gemini's control is no longer a placebo.** A message sent at effort `high`
produced a full reasoning stream — `reasoning_start`, eight `reasoning_delta`
events over ~14s, `reasoning_end` — and the model's actual thinking is stored on
the message.

That last part is the load-bearing evidence, because `includeThoughts` is only set
when thinking is on: Google returns no thought parts without it. Seeing them back
means the whole chain held — the picked tier reached the resolver, the catalog's
`reasoning_dialect: tier` routed it to `thinkingLevel` rather than
`thinkingBudget`, the field reached the request, and Google accepted the shape.
Before this PR that path returned nil and sent nothing at all.

Not covered: the budget dialect (Gemini 2.5 has been retired upstream, so there is
no model left to exercise it), and Anthropic's explicit `disabled` path.

## Added after live testing: unknown Claude models get the adaptive wire

Testing this stack against a Vercel-gateway model surfaced a gap: a Claude added outside our curated templates has no `thinking_mode` and no effort tiers, so the empty-mode bridge resolved it to `toggle` — which on the Anthropic wire means the legacy `thinking{enabled, budget_tokens}` shape, and Claude 4.6+ rejects that outright:

```
"thinking.type.enabled" is not supported for this model.
Use "thinking.type.adaptive" and "output_config.effort" to control thinking behavior.
```

So reasoning was effectively broken for any Claude the preset list didn't know about, and the list can't keep up with releases by construction.

The empty-mode bridge now reads the model id before falling back: parses as Claude ≥4.6 → adaptive; recognizably Claude but no readable version → adaptive too (unknown Claude ids skew new, and the wrong wire is a hard 400 on new models but merely inert on old ones); pre-4.6 Claude and everything non-Claude keep the old toggle fallback. A declared `thinking_mode` still beats all of it in either direction. Cline, LiteLLM and OpenCode each converged on this same default independently; none of the surveyed projects retries a 400 with a different shape, and neither do we.

Verified live: the exact gateway model that produced the 400 above (`anthropic/claude-sonnet-5`, zero metadata) now sends `thinking{type:"adaptive"}` and completes.

## Post-QA merge-readiness fixes

The human QA above was performed on the earlier head. Commit `756243be7` preserves the exercised Gemini tier path and adds regression coverage for legacy Google rows, always-on models, trusted catalog re-import, managed capability revocation, and dormant settings. Agent-run Docker acceptance used a fresh PostgreSQL schema and the current server, including an old-row to catalog-reimport upgrade. The refreshed stable Google catalog entries and MiniMax-M2.7 declaration have not been walked by a human.


chen-ran (approved):
Reviewed current head 756243be7 after #992 landed. Legacy Google rows fail closed, trusted re-import restores wire metadata, managed catalogs revoke stale capabilities, and always-on models expose no false controls. Focused race, full Go/Web, and isolated Docker/PostgreSQL upgrade acceptance pass.

PR #994, opened 2026-08-11
feat(capabilities): read reasoning capabilities from models.dev and OpenRouter

HoneyBBQ:
Stacked on #993. The data-source half of #982's third step.

Adds the sources and the merge. Wiring synccaps to them is deliberately separate —
this PR can be reviewed on whether the data is read correctly, without also
arguing about how much of the catalog discovery should own.

## Why LiteLLM cannot answer our question

It is a pricing table. It reports which OpenAI-wire flags a model happens to set
(`supports_reasoning`, `supports_xhigh_reasoning_effort`) and has no vocabulary
for the shape a provider actually speaks. That is why the templates carry
`reasoning_dialect`, the budget bounds and the disable token by hand, and why
synccaps needs two rules to stop itself erasing them.

**A correction to my own earlier claim:** I first compared coverage across all 603
models and concluded LiteLLM covered Anthropic, OpenAI and xAI better. Everything
in that margin was an old non-reasoning model — `claude-3-haiku`,
`gpt-3.5-turbo-0125`, `gpt-4-0613` — which LiteLLM keeps because it prices them
and models.dev curates away because it describes capabilities. Measured against
the 167 models our templates mark as reasoning, models.dev knows 125 and is ahead
everywhere the two overlap.

## What each source is for

| | models.dev | OpenRouter |
|---|---|---|
| wire dialect (tier vs budget) | ✅ only source | — |
| budget bounds | ✅ only source | — |
| tier list | ✅ authoritative | ✅ |
| **can it be turned off** | implied by option shape | ✅ `mandatory`, direct |
| **does omission mean off** | ✗ cannot express | ✅ `default_enabled` |

OpenRouter is a gateway rather than a description, so it *has* to know which
models refuse to stop thinking and which think unless told otherwise — both fall
out of its own routing code rather than from curation.

`Merge` takes each source for what it alone knows rather than ranking them. Where
they overlap, **a disagreement is reported rather than resolved silently** —
telling us something is wrong is the main thing a second source is for.

Checked against both live catalogs on six models spanning every case: they agree
everywhere, zero conflicts, and `claude-opus-4-6` picks up `default_enabled:false`
matching Anthropic's own per-model table.

## Three things the real catalog taught

Each is now a test, and each was a bug first:

1. **Provider entries are overlays.** They carry only what differs and inherit the
   rest via `base_model` from a top-level `models/` tree I had not read. Of 2994
   overlays, 2751 have no reasoning flag of their own, 2255 no modalities, 1608 no
   context limit — so gpt-5.2, claude-sonnet-4-5 and every Gemini 2.5 resolved as
   unknown while the data sat one file away.
2. **The option list must be read even without the reasoning flag** — 1631 of 2737
   entries carry options and no flag.
3. **Accepted is not preferred.** Claude 4.6 lists both `effort` and
   `budget_tokens` because the API takes both, but budget is deprecated there and
   rejected on 4.7+. A "budget wins" rule would have downgraded a correct
   hand-written dialect, so tiers take precedence when present.

## Data model gaps this closed

- **`always` is its own thinking mode.** A model that reasons and takes no
  direction was stored as a toggle with an empty tier list and off support
  "rejected" — which reads as "switchable, but refuses to switch".
- **`reasoning_default_on` is separate from off-ability.** Claude 4.6 can be turned
  off *and* omitting the field means off; Opus 5 can be turned off but thinks by
  default, so omitting leaves thinking running — billed, and invisible to a user
  who believes it is off. The adaptor now sends an explicit `disabled` only where
  both facts call for it.

Whether a model *accepts* an explicit disable — which splits Anthropic's catalog
three ways and only its documentation states — stays hand-declared and still wins
over both sources.

### DIFF
--- a/internal/agent/application/service.go
+++ b/internal/agent/application/service.go
@@ -723,6 +723,7 @@ func (s *Service) buildBaseRunConfig(ctx context.Context, p baseRunConfigParams)
 		ReasoningConfig:       reasoningConfig,
 		ReasoningDialect:      chatModel.Config.ReasoningDialect,
 		ReasoningOffSupport:   chatModel.Config.ReasoningOffSupport,
+		ReasoningDefaultOn:    chatModel.Config.ReasoningDefaultOn,
 		ThinkingBudgetMin:     chatModel.Config.ThinkingBudgetMin,
 		ThinkingBudgetMax:     chatModel.Config.ThinkingBudgetMax,
 	})
--- a/internal/agent/tool/subagent.go
+++ b/internal/agent/tool/subagent.go
@@ -1643,6 +1643,7 @@ func (p *SpawnProvider) resolveModel(
 		ReasoningConfig:       reasoningConfig,
 		ReasoningDialect:      modelInfo.Config.ReasoningDialect,
 		ReasoningOffSupport:   modelInfo.Config.ReasoningOffSupport,
+		ReasoningDefaultOn:    modelInfo.Config.ReasoningDefaultOn,
 		ThinkingBudgetMin:     modelInfo.Config.ThinkingBudgetMin,
 		ThinkingBudgetMax:     modelInfo.Config.ThinkingBudgetMax,
 	})
--- a/internal/capabilities/discover.go
+++ b/internal/capabilities/discover.go
@@ -51,6 +51,9 @@ type Capabilities struct {
 	ThinkingBudgetMin   *int
 	ThinkingBudgetMax   *int
 	ReasoningOffSupport string
+	// ReasoningDefaultOn reports whether omitting the thinking field leaves the
+	// model thinking. nil when no source said.
+	ReasoningDefaultOn *bool
 }
 
 // effortOrder is the canonical low→high ordering used to render effort lists.
--- a/internal/capabilities/discover.go
+++ b/internal/capabilities/discover.go
@@ -43,6 +43,14 @@ type Capabilities struct {
 	FileInput *bool
 	// ContextWindow is nil when unknown.
 	ContextWindow *int
+	// ReasoningDialect, the budget bounds and ReasoningOffSupport are wire facts
+	// LiteLLM could not express at all: it reported which OpenAI-wire flags a model
+	// set, not which shape the provider speaks. models.dev models the shape
+	// directly, so discovery can now fill what the templates used to carry by hand.
+	ReasoningDialect    string
+	ThinkingBudgetMin   *int
+	ThinkingBudgetMax   *int
+	ReasoningOffSupport string
 }
 
 // effortOrder is the canonical low→high ordering used to render effort lists.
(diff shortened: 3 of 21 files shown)


HoneyBBQ:
Closing. This was a capability upgrade rather than a bug fix, and the line of work
had drifted well past what anyone reported.

The review fixes that were mixed into it have moved to #993, where they belong:
`reasoning_off_support` now distinguishes the four ways an Anthropic model answers
an explicit disable, `reasoning_default_on` separates "can it be turned off" from
"does omitting the field leave it thinking", and `always` becomes its own thinking
mode. Those are corrections to a design already under review; the source swap was
not.

## What the research established, for whoever picks this up

Worth keeping, because the conclusions were not obvious and two of them reversed
my own earlier claims:

- **models.dev is the only source that models a provider's control as a typed
  shape** (`toggle` / `effort{values}` / `budget_tokens{min,max}`), which is what
  `reasoning_dialect` and the budget bounds need. LiteLLM reports which OpenAI-wire
  flags a model set and cannot express the shape at all.
- **My first coverage comparison was wrong.** I measured across all 603 models and
  concluded LiteLLM covered Anthropic and OpenAI better; everything in that margin
  was an old non-reasoning model it prices and models.dev curates away. Measured
  against models that actually reason, models.dev leads wherever they overlap.
- **OpenRouter's `mandatory` and `default_enabled` are unique to it** and cover all
  406 of its models, not just Anthropic. `default_enabled` is the only machine-
  readable source for the omission hazard above.
- **Anthropic's own per-model table is machine-readable** at
  `platform.claude.com/docs/en/build-with-claude/thinking-troubleshooting.md`
  (registered in their `llms.txt`, so it is an intended consumption path). It has a
  literal "Rejected with 400" column and is the only source covering all four
  off-support cases. That is the highest-value remaining source.
- **Do not add Vercel AI Gateway.** It mirrors models.dev but is wrong on the two
  models that matter: Fable 5 marked `toggle` (it 400s), Opus 5 `null`.
- **Do not restructure the per-provider templates.** I measured 1.30x redundancy,
  and 66% of the apparent duplication is genuine divergence — `claude-sonnet-4-5`
  is `[low, medium, high]` direct and `[disable]` through OpenRouter, because a
  gateway offers controls a vendor does not. Cline consumes models.dev at 2.3x
  redundancy and deliberately discards its inheritance; models.dev's own guidance
  exempts first-party hosts from `base_model`. The real risk is drift, not
  duplication — Roo Code has the same model at 32000 and 8192 max tokens in two
  files, and LiteLLM ships a silent failure where thinking is dropped for exactly
  the Claude models its code targets.
- **MiniMax M2.x returns 200 and keeps thinking** when sent an explicit disable —
  worse than a 400, since the bill is the only signal. A `silently_ignored` case
  would be needed before anyone declares those models disableable.

Branch `feat/synccaps-modelsdev` keeps the source readers and their tests if this
is picked up later.

PR #976, opened 2026-08-07
refactor(web): rewrite sidebar Recents pagination without virtual list

qqqqqf-q:
## Summary

- Remove TanStack Virtual from sidebar Recents and folder session lists; render cursor-paginated sessions with native `v-for` plus sentinel-based load-more and viewport prefetch via `use-sidebar-infinite-scroll`.
- Unify folder expanded sessions with the same infinite-scroll path (`folder-sessions-list.vue`) instead of a separate Show more button.
- Harden session bootstrap: retry session fetch when `initialize()` fails, toast on final failure, and keep sidebar skeleton/shimmer tuned for list loading.

## Test plan

- [x] Open a bot with 100+ sessions; scroll Recents — load-more triggers without scroll jumps (215 sessions on qf 2)
- [ ] Expand a folder with many sessions — same sentinel/spinner behavior as Recents
- [ ] Switch bots — list resets scroll top and reloads from first page
- [ ] Simulate failed session bootstrap — retry succeeds or error toast appears
- [ ] Mobile sidebar panel — Recents/Folders behave consistently
- [x] `pnpm exec vitest run apps/web/src/components/sidebar/use-sidebar-infinite-scroll.test.ts`

Human QA during development: Recents load-more / viewport prefetch, skeleton shape and shimmer speed feedback incorporated.

PR #995, opened 2026-08-12
refactor(web): schedule page follows the settings page language; shared session picker

sheepbox8646:
## Summary

The schedule tab's task list was the last schedule surface still off the design contract. This brings it onto the settings page language and fixes two information gaps in the rows.

Diagnosed against `packages/ui/skills/web/SKILL.md` + its dirty→clean table; rebuilt by mirroring `bot-connectors.vue`, the clean same-shape sibling in the same directory.

## What was off-contract

| Dirty | Rule |
|---|---|
| `border-dashed` empty state | dashed is reserved for the "+ Add another" tile beside real items; an empty surface keeps the populated frame |
| `<Calendar>` glyph stacked in that box, `text-muted-foreground/40` | decorative icon in a card + hand-mixed alpha |
| hand-rolled empty state | the `Empty` owner exists |
| two-column grid of hand-rolled clickable cards | a list of manageable rows is `SettingsSection` + `SettingsRow`, not an object-picker grid |
| `hover:bg-accent/30 dark:hover:bg-accent` | hand-mixed alpha + a `dark:` override — the tell of starting from a raw color; the overlay ladder is scheme-agnostic |
| hand-rolled `Dialog` + two buttons for delete | `ConfirmDeleteDialog` owner |
| `class="text-muted-foreground"` injected onto a `<Button>` | the className red line |

## What changed

**List** — one titleless `SettingsSection` (the page title already names the concern) of `SettingsRow`s, each with a `Switch` and an `icon-sm` row menu. `Empty` inside that card (borderless, no icon tile, title + description + one guiding action), `InlineLoadingRow surface="card-row"` for loading, `ConfirmDeleteDialog` for delete. Box count drops from N cards to one card of rows.

**Rows carry the schedule facts.** The cron description used to disappear whenever a task had its own description — on a page whose subject is *when things run*. And the next fire time was only ever visible in the sidebar. Both now ride the row's single line (`description · every day at 09:00 · Next run Tomorrow 9:00 AM`), using the shared locale-aware `formatCalendarTime`. A disabled task promises no next run, so that clause drops.

**Sidebar card split out.** `schedule-list-item.vue` served both the sidebar panel and the settings tab through a `variant` prop — the same-shape-different-relationship fork that pulled the settings tab off-language. It becomes `components/sidebar/schedule-item.vue`, stating one shape, with its hover on the sidebar's rung of the neutral overlay ladder (the token `session-item.vue` and `folders-section.vue` already use). No visual change to the sidebar.

**Editor** — Run Limit joins the form column it lives in as a `FieldStack` instead of a squat `sm` side-by-side row; the placeholder carries the empty-value meaning so no help line repeats it.

## Deliberately dropped

Whole-row click-to-edit on the settings tab. Keeping it beside the row's `Switch` needs either a hand-rolled row (guard-flagged) or a new `SettingsRow` variant in the `packages/ui` submodule; Edit is the first item of the row menu. Say the word and I'll add it back via the stay-local stretched-overlay pattern `ui-owners` sanctions for exactly this shape.

## Shared session picker (commit 2)

`components/session-select/` composes the existing `SearchableSelectPopover` into the three things the sidebar list already gives you, so surfaces that must name an existing session stop hand-rolling their own list:

- **Folder groups** — unfiled sessions first, headerless until a folder exists (a bot with no folders reads as one plain list), then one group per live folder in the sidebar's order. A session bound to a folder that is gone keeps its own group rather than silently joining the unfiled bucket.
- **Search** over title, folder name and id.
- **Per-kind marks** — a yellow clock for schedule sessions, the agent mark for ACP, nothing for a plain model chat, with a reserved slot so labels stay aligned. Schedule wins over the ACP mark here, unlike the sidebar row: a picker that exists to separate schedule-owned sessions from ordinary chats should say so even when that session runs an agent.

The grouping is a pure function in `options.ts`, so those rules are covered by 10 unit tests instead of asserted by eye. A `modes` prop lets a caller list only the session modes it can accept.

`ModelSelect` also gains a `reasoningOptions` passthrough to `ModelOptions`, for callers whose effort tiers come from a runtime rather than the model's own capabilities.

> **Both land ahead of their first consumer.** They were written for the schedule editor's execution parameters, which are not on `main` — `main`'s `schedule` table has no `run_target`/`model_id`/`reasoning_effort` columns and `ScheduleSchedule` has no such fields. The picker rewrite that consumes them (`schedule-execution-fields.vue`) and the backend change that accepts an explicit "Off" effort (`internal/schedule/execution.go`) have no file to modify here; they are held back until that feature reaches `main`.

## Testing

- `pnpm build`, ESLint, `node scripts/check-ui-contract.mjs` — no new violations from these files (the guard's `recents.vue` px failure is pre-existing on `main`, from #976).
- `vitest`: 1063 pass (including 10 new `session-select` grouping tests). Three pre-existing failures on clean `main`, verified by re-running them with this branch stashed: `useMediaGallery.test.ts` (`localStorage.getItem is not a function`), `panel-preview.test.ts` (vue-i18n mock missing `createI18n`), and a flaky 15s timeout in `model-list-row.test.ts` that passes standalone.
- Raw-color / `dark:` / inline-style grep clean across the touched files; the two new i18n keys are in en, zh **and** ja.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/web/src/components/session-select/index.vue
+++ b/apps/web/src/components/session-select/index.vue
@@ -1,0 +1,179 @@
+<template>
+  <SearchableSelectPopover
+    v-model="selected"
+    :options="options"
+    :placeholder="resolvedPlaceholder"
+    :aria-label="resolvedPlaceholder"
+    :search-placeholder="t('chat.searchSessionPlaceholder')"
+    :search-aria-label="t('chat.searchSessions')"
+    :empty-text="emptyText"
+    popover-class="w-[var(--reka-popover-trigger-width)]"
+  >
+    <!-- Custom trigger: the default one is text-only, and the selected
+         session's mark (schedule clock / agent) has to survive collapse —
+         otherwise the kind is visible only while the menu is open. -->
+    <template #trigger="{ open }">
+      <button
+        data-slot="select-trigger"
+        data-size="default"
+        :data-placeholder="triggerLabel ? undefined : ''"
+        type="button"
+        :aria-expanded="open"
+        :aria-label="resolvedPlaceholder"
+        :class="[selectTriggerClass, 'w-full']"
+      >
+        <span class="flex min-w-0 items-center gap-2">
+          <SessionKindIcon
+            v-if="selectedMark"
+            :kind="selectedMark.kind"
+            :agent-id="selectedMark.agentId"
+            :label="selectedMark.label"
+          />
+          <span class="line-clamp-1">{{ triggerLabel || resolvedPlaceholder }}</span>
+        </span>
+        <ChevronsUpDown class="opacity-50" />
+      </button>
+    </template>
+
+    <template #option-icon="{ option }">
+      <SessionKindIcon
+        :kind="markOf(option).kind"
+        :agent-id="markOf(option).agentId"
+        :label="markOf(option).label"
+        reserve-space
+      />
+    </template>
+  </SearchableSelectPopover>
+</template>
+
+<script setup lang="ts">
+import { computed, ref, watch } from 'vue'
+import { useI18n } from 'vue-i18n'
+import { ChevronsUpDown } from 'lucide-vue-next'
+import { selectTriggerClass } from '@felinic/ui'
+import { getBotsByBotIdSessions } from '@memohai/sdk'
+import type { SessionSession } from '@memohai/sdk'
+import SearchableSelectPopover from '@/components/searchable-select-popover/index.vue'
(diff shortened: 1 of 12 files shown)


sheepbox8646:
Superseded by #997, which is based on `main`. Both commits from this PR are contained there (`refactor(web): schedule page…` and `feat(web): shared session picker…`), alongside the rest of #962's work replayed onto `main`.

PR #1009, opened 2026-08-13
fix(agent): cap replayed reasoning at the newest assistant turn

### LINKED ISSUE #1008: [Bug]: replayed reasoning grows without bound across a conversation
Replayed reasoning grows without bound. Every turn sends back every earlier turn's reasoning blocks, and nothing ever caps it.

Climbing in a proxy log against a live Claude 4.6:

```
messages=17  replayed 13 blocks
messages=19  replayed 19 blocks
messages=21  replayed 19 blocks
messages=23  replayed 26 blocks
messages=25  replayed 33 blocks
```

Measured on that same conversation:

```
redacted blocks: 35   (14,352 bytes of ciphertext)
signed blocks:    5
```

None of it is cached, either — `prompt_cache.go` only puts breakpoints on the system prompt and the last tool definition, nothing in the messages array. So those 14 KB are re-sent and re-billed at full price on every single turn, and the bill grows with the conversation.

## Why it got noticeable recently

The accumulation itself is old — we've never had a ceiling. What changed is how much there is to accumulate.

`redacted_thinking` blocks used to be dropped on parse (the bug twilight-ai#45 fixed), so they never entered history. Now that they're handled correctly they do, and they're 87% of the total. So this isn't a regression we introduced; it's a missing ceiling that a correctness fix made impossible to ignore. Going back isn't an option — dropping redacted blocks is the documented top cause of `thinking or redacted_thinking blocks in the latest assistant message cannot be modified`.

## Why the existing stripping doesn't catch it

`stripToolMessages` only rewrites an assistant message when `hasToolCallContent(m)` is true (`internal/agent/application/service_history_messages.go`). A plain conversational turn has no tool call, so it never enters that path and keeps its reasoning forever. Confirmed with a probe: two plain turns, both keep their blocks.

## What to do

Cap it at the newest turn. Anthropic verifies the thinking blocks of the latest assistant message and filters older turns server-side ("You don't need to prune old thinking yourself"), so that's the only one that has to go back — and this has to apply to every assistant message, not just the ones carrying a tool call.

## Not the cause of the timeouts

I initially thought the growing request was what tripped our 2-minute provider timeout, since the two showed up together. That's wrong, and worth writing down so nobody chases it: after capping the replay to 2 blocks the timeouts still happen at the same rate. Tracked separately in #1010.

HoneyBBQ:
Closes #1008

Replayed reasoning had no ceiling. Every turn sent back every earlier turn's blocks, so a conversation kept growing its own request:

```
messages=17  replayed 13 blocks
messages=23  replayed 26 blocks
messages=25  replayed 33 blocks
```

And none of it is cached — `prompt_cache.go` only breakpoints the system prompt and the last tool definition, nothing in the messages array — so all 14 KB of ciphertext measured on that conversation was re-billed at full price every turn.

The accumulation is old; what changed is the volume. `redacted_thinking` blocks used to be dropped on parse, so they never entered history. twilight-ai#45 fixed that and now they're 87% of the total (35 redacted vs 5 signed). Not a regression, then — a missing ceiling that a correctness fix made impossible to ignore. Reverting isn't on the table: dropping redacted blocks is the documented top cause of `thinking or redacted_thinking blocks in the latest assistant message cannot be modified`.

Anthropic tells us where to draw the line — it verifies the latest assistant message's thinking sequence and filters older turns server-side ("You don't need to prune old thinking yourself"). So the newest turn keeps its blocks in full, empty-text redacted ones included, and everything older loses them.

The catch, and why this took two commits: `stripToolMessages` only rewrites an assistant message when `hasToolCallContent(m)` is true. A plain conversational turn never enters that path, so the first commit capped nothing for most messages — a probe over two plain turns showed both keeping their reasoning. The second drops reasoning from every older assistant message regardless of tool calls. A turn whose only content is reasoning keeps its original shape, since emptying it would leave a contentless assistant message that providers reject.

## Verified live

Ran it against a live Claude 4.6 through a logging proxy. Replayed block counts went from climbing (13 → 19 → 26 → 33) to tracking just the newest turn (6 → 2 → 2), with one turn replaying nothing at all because its latest assistant message had no reasoning. At `messages=24` that's 2 blocks where the old behaviour sent 33. No 400s.

Seven guards, including one asserting the count stops growing across eight turns. I reverted the widening to watch it fail (`got 16, want 2`) — a dropped signature is silent until a provider rejects the next request, so a guard that has never failed isn't worth much here.

## One correction

An earlier version of this description blamed the growing request for the 2-minute provider timeouts we were seeing, because they showed up together. That was wrong: after this change the requests carry 2 blocks instead of 33 and the timeouts happen just the same. Filed separately as #1010.

twilight-ai#45 fixed per-block replay upstream; the two are independent and can merge in either order.

### DIFF
--- a/internal/agent/application/service_history_compaction.go
+++ b/internal/agent/application/service_history_compaction.go
@@ -64,8 +64,9 @@ func historySourceMessageIDsForMessages(messages []ModelMessage, records []histo
 		if strings.TrimSpace(record.DBMessageID) == "" {
 			continue
 		}
-		key := modelMessageSourceKey(record.ModelMessage)
-		positions[key] = append(positions[key], i)
+		for _, key := range modelMessageSourceKeys(record.ModelMessage) {
+			positions[key] = append(positions[key], i)
+		}
 	}
 	cursors := make(map[string]int, len(positions))
 	lastMatched := -1
@@ -90,6 +91,22 @@ func historySourceMessageIDsForMessages(messages []ModelMessage, records []histo
 
 func modelMessageSourceKey(message ModelMessage) string {
 	return strings.ToLower(strings.TrimSpace(message.Role)) + "\x00" + string(message.Content)
+}
+
+// modelMessageSourceKeys returns every key a record's message can be recognised
+// by. Context assembly caps replayed reasoning, so a durable assistant turn
+// reaches the model with its reasoning dropped or projected to text; indexing
+// the record under that shape as well keeps the message resolvable to its DB row
+// instead of silently losing provenance for fork sources and context accounting.
+func modelMessageSourceKeys(message ModelMessage) []string {
+	keys := []string{modelMessageSourceKey(message)}
+	if !strings.EqualFold(strings.TrimSpace(message.Role), "assistant") {
+		return keys
+	}
+	if stripped := dropReasoning(message); string(stripped.Content) != string(message.Content) {
+		keys = append(keys, modelMessageSourceKey(stripped))
+	}
+	return keys
 }
 
 func stripToolMessagesWhenCompactionSummaryIsActive(messages []ModelMessage, records []historyfrag.HistoryRecord) []ModelMessage {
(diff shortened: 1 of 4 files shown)


chen-ran (changes requested):
- 普通 Web/Channel 路径会提前丢掉全部 reasoning，这次修复只对 fallback 生效。
- 切换模型或 provider 时，旧模型的 reasoning 签名仍会发给新模型，可能触发 400 或泄露隐藏上下文。
- 最新 tool-use turn 会保留 reasoning 却删除对应的 tool_use，破坏 Anthropic 要求的完整回放。

chen-ran (approved):

PR #1041, opened 2026-08-17
feat(context): hook authority split and provider-attempt preflight

### LINKED ISSUE #890: Make system prompt assembly budget-aware
## Problem

The native runtime builds the system prompt in several passes:

- `GenerateSystemPrompt` joins the common template with a mode template and injects bot info, skills, platform identities, and workspace instruction files.
- Tool usage is inserted later, after tool registration.
- Prompt hooks can append more text after the prompt has been built.

Only workspace files have a local size cap. There is no budget for the complete system prompt, so the final size can grow with installed skills, registered tools, identities, workspace files, and hook output. The history budget also does not account for this cost.

The current typed context model already has section kinds, priorities, cache classes, and `BudgetPolicy`, but system prompt rendering does not enforce those fields. Prompt assembly still depends on string concatenation and heading-based insertion.

## Proposed change

Build the system prompt as an ordered set of named sections before rendering it to a string. Each section should declare:

- whether it is protected or optional;
- its priority and size limit;
- what to do on overflow;
- whether it is stable enough for prompt caching;
- its source, so pruning is visible in logs and the context manifest.

Use one total system prompt budget derived from the selected model's context window, with explicit reserves for the current request, tool schemas, and model output. Keep a conservative fallback for providers without tokenizer support.

The composer should also shorten the static common and mode templates. Remove repeated instructions, render mode-specific text only for the active mode, and include capability guidance only when the capability is registered. Keep stable text ahead of per-turn content where the provider's cache behavior allows it.

When the prompt exceeds its budget, prune it in a deterministic order. Start with low-priority dynamic content and trim within a section before dropping the whole section. Protected safety rules, instruction priority, and the active mode contract must not disappear silently. If protected content alone cannot fit, return a clear error instead of sending a broken prompt.

## Scope

Apply the composer to every native session mode:

- chat
- discuss
- heartbeat
- schedule
- subagent

It must cover the common and mode templates, bot identity, platform identities, skill catalog, registered tool usage, workspace instruction files, and prompt hook output.

Conversation history compaction is out of scope. This issue is about the system prompt and the budget it consumes.

## Acceptance criteria

- The rendered system prompt stays within its computed budget for every native session mode.
- Static prompt size has a checked-in baseline and a regression limit. The implementation should reduce the current baseline, not only cap future growth.
- Required policy and mode sections remain intact after pruning.
- Guidance for an unavailable tool or capability is never rendered.
- Large skill catalogs, tool usage blocks, identity lists, workspace files, and hook output are pruned predictably and leave an explicit marker.
- Section order is deterministic and keeps the stable prefix cache-friendly.
- Logs or the context manifest report total prompt cost, per-section cost, and every prune decision without recording private prompt contents.
- Tests cover each session mode, oversized dynamic sections, Unicode input, missing tokenizer support, and the protected-content overflow path.

ChrAlpha:
## What changes

Hook context gains a typed authority seam: hooks materialize bounded system sections (at most 16 per run, each attributed to its hook and ordered deterministically), contextview governs them as first-class system fragments instead of splicing text into the static prompt, and legacy hook context stays isolated on its own path. Discuss runs keep hook sections and context slices isolated per mode entry, and hook tool output respects an explicit default output cap.

Provider attempts gain a unified preflight: every provider-loop step re-enters the budget/selection path before it is sent, with a step reselector installed per provider, serialized step envelopes charged at their true marginal cost, raw retry input preserved, and step truncation tiers unified under `internal/agent/context/limit`. The mutation ledger records provider-attempt audits — model and client type, per-step content-light snapshots, cache usage tiers, and the loop selection mode — and lifecycle snapshots carry them without exposing payloads. Injected step messages and read-media injections are audited, retry budget errors are suppressed where retries own recovery, and provider cache usage aggregates across tiers.

## Scope

This PR contains the hook system-section materialization and governance, the provider-attempt preflight and step reselection, unified step truncation tiers, attempt audit ledger and snapshot plumbing, cache plan consumption, and the provider-loop test split.

It does not add lifecycle aggregation, budget-failure/fallback classification, tool-roster attribution, or web consumers (PR8), and it does not change prompt text or baselines (PR7).

This is part 4 of the landing work motivated by #890. It builds on the compiler (#959), section policy (#968), budget plan (#1012, landed via #1039), decision fencing (#960), and lifecycle persistence (#969), but does not close #890.

## Restack notes

The 29 PR4 commits plus two adaptation commits were replayed from the immutable PR3 head (`f1525fa16`) onto upstream main `dc2be8804` (v0.18.0). Notable resolutions:

- `feat(hooks): materialize system sections` — #938 (Registry Skill Packages) removed the plugin-bundle hook source (`sourceKindPlugin` and plugin-aware output limits) after this commit was written. The resolution keeps #938's single-source model: hook/system-section attribution stays, the plugin metadata block is dropped, and `hookMaxOutputBytes(cfg)` keeps its post-#938 single-argument shape with this PR's explicit default-cap argument.
- `test(agent): align restacked tests with upstream seams` — #969 moved the fragment lifecycle tests into an external test package and #991/#1022 changed `buildGenerateOptions` to take a context and prepare-step hook; the affected tests are qualified and call the merged signature.
- `fix(context): keep empty attempt audits nil in snapshots` — #969's snapshot assertions compare with `reflect.DeepEqual` after a JSON round trip, which distinguishes nil from empty slices; empty step and cache-usage audits now stay nil.
- `test(hooks): scope section caps to the post-#938 source model` — the narrowest-cap test exercised per-plugin output limits that #938 removed; it now pins the aggregate section cap against `Defaults.MaxOutputBytes` alone.
- `fix(bridge): tolerate exec streams closed before stdin lands` — CI surfaced a latent bridge-client race this PR's stdin-bearing hook execs hit reliably on slow runners: a command that exits without consuming stdin closes the exec stream while the stdin `Send` is in flight, and gRPC reports that close as a bare `io.EOF`, failing an exec that actually succeeded. Half-closed sends now fall through to the receive loop, which carries the authoritative result or status. Pinned by a deterministic flow-control regression test.

## Verification

The following checks passed on final head `e8dfa60ef`:

- `go build ./...`
- `go test -count=1 ./...` (zero failures)
- `go test -race -count=1` over application, native, contextview, hooks, context/fragment, and workspace/bridge
- `golangci-lint run` over the touched Go packages
- `git diff --check`; merge-base equals the upstream main head, 33 commits, zero merge commits
- the native package re-run three times at `GOMAXPROCS=2` after the bridge fix (the CI failure mode no longer reproduces)

## Landing stack

| PR | Scope |
| --- | --- |
| PR1 (merged #959) | Fragments-first authoritative compiler with byte-equivalent provider output |
| PR2 (merged #968) | System-section retention, drop policy, grouped rendering, and capability gating |
| PR3 (merged #1012 via #1039) | Unified `ContextBudgetPlan` and budget-aware behavior for all five modes |
| PR4 (this PR) | Hook authority split and provider-attempt preflight |
| PR5 (merged #960) | Runtime decision correctness and fencing for #865 |
| PR6 (merged #969) | Run-keyed lifecycle persistence and read ownership |
| PR7 | Prompt shrink, baselines, and repository-level gates |
| PR8 | Cross-stack lifecycle, budget, fallback, and tool-roster integration seams |


ChrAlpha:
The CI `Test` failures (`before model call hook failed: EOF`, a different hook test each run) were a latent bridge-client race, not a hook bug: a command that exits without consuming stdin closes the exec stream while the client's stdin `Send` is in flight, and gRPC reports that close as a bare `io.EOF` — failing an exec that actually succeeded. This PR's before-model-call hooks are the first stdin-bearing execs on that path, so slow runners hit the window reliably while local runs never did. Fixed in `e8dfa60ef`: half-closed sends fall through to the receive loop, which carries the authoritative result or status; a flow-control-based regression test reproduces the race deterministically.

ChrAlpha:
The bridge gRPC exec stdin EOF fix (e8dfa60ef, `internal/workspace/bridge/`) is a pre-existing main bug — extracted to standalone #1048. Once that merges, this branch will be rebased so the bridge diff drops out of this PR; reviewers can skip those two files here.

sheepbox8646 (approved):

PR #1048, opened 2026-08-18
fix(bridge): tolerate exec streams closed before stdin lands

ChrAlpha:
## Problem

`Client.ExecWithOptions` returns any error from the config `Send`, stdin `Send`, or `CloseSend` as-is. When the executed command exits without consuming its stdin, the bridge server handler returns and closes the stream while those sends are still in flight, and gRPC reports the half-closed stream as a bare `io.EOF`. The exec actually succeeded, but the client reports it as a failure.

Per the gRPC streaming contract, `io.EOF` from `SendMsg` means "stream closed by the peer — call `RecvMsg` for the authoritative status". The receive loop below already carries the real result, so the fix tolerates `io.EOF` on the three sends and falls through to it.

## Impact on main

This is a pre-existing bug, not specific to any feature branch. On main the stdin-bearing path is `internal/hooks/service.go` (`ExecWithStdinEnv`): bot lifecycle hooks pipe their JSON payload via stdin, so a hook command that exits without draining stdin can have a successful run reported as an `EOF` error — for `PreToolUse` that wrongly blocks the tool call. The window is timing-dependent, which is why it only shows up on slow runners (reproduced consistently on 2-core CI, never locally).

## Regression test

The test makes the race deterministic instead of timing-dependent: the server responds and returns after the config message without draining stdin, and the client sends an 8MB stdin payload. HTTP/2 flow control (64KB window, no application reads means no window updates) guarantees the stdin `Send` is still blocked when the server closes the stream, so the send observes `io.EOF` on every run.

## Relation to #1041

Extracted from #1041 (commit e8dfa60ef there), where the new before-model-call hook path made the race show up on every CI run. Landing this first keeps that CI green; #1041 will be rebased afterwards so its bridge diff disappears.

### DIFF
--- a/internal/workspace/bridge/client.go
+++ b/internal/workspace/bridge/client.go
@@ -279,7 +279,10 @@ func (c *Client) ExecWithOptions(ctx context.Context, command, workDir string, t
 		return nil, mapError(err)
 	}
 
-	// Send config message first
+	// A command may exit without consuming its stdin; the server then closes
+	// the stream while these sends are still in flight, and gRPC reports the
+	// close as io.EOF. The receive loop below carries the authoritative
+	// result or status, so half-closed sends fall through to it.
 	err = stream.Send(&pb.ExecInput{
 		Command:        command,
 		WorkDir:        workDir,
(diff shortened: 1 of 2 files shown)


ChrAlpha:
Superseded: #1041 merged with the bridge commit included (landed on main as part of 34a47bf8d), so main already has this exact fix and regression test. Closing as redundant.

PR #1059, opened 2026-08-20
test(agent): fix step budget retry cancellation race

Fodesu:
## Summary

- Fix the mid-stream retry cancellation test context double.
- Model cancellation becoming visible between the stream loop guard and ErrorPart handling.
- Ensure raw provider cancellation errors remain suppressed.

## Validation

- `go test ./internal/agent/runtime/native -run '^TestRunMidStreamRetrySuppressesRawErrorAfterStepBudgetCancellation$' -count=100 -race`
- `go test ./internal/agent/runtime/native`

### DIFF
--- a/internal/agent/runtime/native/provider_retry_budget_error_test.go
+++ b/internal/agent/runtime/native/provider_retry_budget_error_test.go
@@ -12,18 +12,23 @@ import (
 	contextfrag "github.com/memohai/memoh/internal/agent/context/fragment"
 )
 
-// hiddenErrContext models cancellation becoming visible after the retry
-// loop's ctx.Err check but before it handles the provider's ErrorPart. Cause
-// still reaches the embedded cancel-cause context, as it does in the race.
-type hiddenErrContext struct {
+// delayedErrContext models cancellation becoming visible after the retry
+// loop's ctx.Err check but before it handles the provider's ErrorPart. The
+// first Err call is the retry preflight and the second is the loop guard;
+// the third call is the ErrorPart budget check.
+type delayedErrContext struct {
 	context.Context
+	errCalls atomic.Int32
 }
 
-func (hiddenErrContext) Err() error {
-	return nil
+func (c *delayedErrContext) Err() error {
+	if c.errCalls.Add(1) <= 2 {
+		return nil
+	}
+	return c.Context.Err()
 }
 
-func (hiddenErrContext) Done() <-chan struct{} {
+func (*delayedErrContext) Done() <-chan struct{} {
 	return nil
 }
 
@@ -31,7 +36,7 @@ func TestRunMidStreamRetrySuppressesRawErrorAfterStepBudgetCancellation(t *testi
 	t.Parallel()
 
 	baseCtx, cancel := context.WithCancelCause(context.Background())
-	streamCtx := hiddenErrContext{Context: baseCtx}
+	streamCtx := &delayedErrContext{Context: baseCtx}
 	var providerCalls atomic.Int32
 	provider := &atomicMockProvider{
 		stream: func(context.Context, sdk.GenerateParams) (*sdk.StreamResult, error) {


chen-ran (approved):

qqqqqf-q:
**中文摘要(给维护者扫读,细节见下方英文):** 问题真实存在——旧测试在 Go 1.26 下 100% 必挂(根因:1.26 的 `context.Cause` 新增了 `Err()` 短路,旧 mock 的 `Err()` 恒 nil;CI 钉 1.25.7 所以一直绿)。修复方向对,但当前实现没做到它自己声称的事:插桩实测 `Err()` 实际被调 7 次,loop guard 是第 6 次(注释说第 2 次),`<=2` 阈值让 guard 直接看到真实取消并 break,ErrorPart 分支的抑制检查从未执行——证伪实验:删掉该检查,新测试照样全绿(旧 mock + 1.25.7 下删了会红),覆盖从 ErrorPart 抑制静默降级为 guard 拦截。建议改用"cancel 后第一次 `Err` 隐藏、其余透传"的写法(代码在下方),已验证:双 Go 版本 100 次 `-race` 全绿,且删掉抑制检查后双版本都红(真覆盖)。

---

Thanks for fixing this — the problem is real, and worth stating precisely in the PR description: **the old test fails 100% of the time on Go 1.26, not as a race/flake.** Root cause: Go 1.26's `context.Cause` now short-circuits on `c.Err() == nil` before consulting the cancelCtx value, while Go 1.25's `Cause` went straight to `Value(&cancelCtxKey)`. The old double's `Err()` always returned nil, so `Cause` always returned nil. CI stays green only because it's pinned to Go 1.25.7 via `go-version-file: go.mod`. (Verified locally: old test 100/100 FAIL on go1.26.0, 20/20 OK under `GOTOOLCHAIN=go1.25.7`.)

The direction of the fix is right (the double must eventually pass through the real cancellation), and I confirmed this PR's version passes 100/100 `-race` on **both** go1.26.0 and go1.25.7. However, the new double doesn't actually model what its comment — and this PR's second bullet — claims:

> The first Err call is the retry preflight and the second is the loop guard; the third call is the ErrorPart budget check.

Instrumenting `Err()` shows it's actually called **7 times** on Go 1.26, and the loop guard is call **#6**, not #2:

```
#1 context.Cause          (retry preflight)          real=false
#2 providerAttemptDispatchAllowed                    real=false
#3 context.Cause          (SDK guard chain)          real=false
#4 context.Cause          (SDK guard chain)          real=false
#5 contextBudgetGuardProvider.DoStream               real=false
   --- mock stream runs, cancel happens ---
#6 runMidStreamRetry      (drain-loop guard)         real=true  → break
#7 context.Cause          (test assertion)           real=true
```

With the `<= 2` threshold, the drain-loop guard (#6) already sees the real cancellation and breaks — **the `ErrorPart` branch's `contextStepBudgetError` suppression never executes.** The test passes via guard interception, not via the path it exists to cover.

Falsification check: I deleted the `contextStepBudgetError` check in the `ErrorPart` branch of `agent.go` — the new test still passes 100% on Go 1.26 (zero coverage of the suppression). The old double + Go 1.25.7 with the same deletion fails with `retry leaked raw provider error event` — i.e. the old test did cover that path, and this PR silently downgrades the coverage to "guard saw cancellation," which the main flow already exercises.

Count-based gating is also inherently fragile: the call count depends on SDK internals and the Go version (7 on 1.26; fewer on 1.25 where `Cause` skips `Err`), so the next upgrade can silently shift which path the test exercises — which is exactly how this test broke in the first place.

Suggested rewrite — hide exactly the first `Err` call after cancellation, pass everything else through:

```go
// delayedErrContext hides cancellation from the first Err call that
// follows the provider stream invocation (the drain-loop guard), then
// reveals it to every later call (the ErrorPart budget check).
type delayedErrContext struct {
	context.Context
	hideOnce atomic.Int32
}

func (c *delayedErrContext) Err() error {
	if c.hideOnce.CompareAndSwap(1, 0) {
		return nil
	}
	return c.Context.Err()
}
```

plus `streamCtx.hideOnce.Store(1)` in the mock stream after `cancel(...)`. Before the stream runs, `Err` passes through (naturally nil, so it's compatible with any number of pre-stream calls and both Go versions); the guard hides exactly once; the `ErrorPart` budget check then sees the real cancellation. This models the exact window the PR describes, decoupled from call counts.

I verified this variant end-to-end:
- 100/100 `-race` pass on both go1.26.0 and go1.25.7;
- with the `ErrorPart` suppression deleted, it **fails** on both versions (`retry leaked raw provider error event`) — i.e. it genuinely covers the suppression path.

Small ask: please also mention the Go 1.26 `context.Cause` change in the PR description and retitle (this is a deterministic failure on 1.26, not a race) — otherwise the next person who sees this test break will go hunting for a flake in the wrong place.

PR #1077, opened 2026-08-24
feat(context): pre-materialization admission boundary + memory scheduling design

chen-ran:
## What this is

Two things, in one reviewable unit:

1. **The design doc** `docs/design/context-memory-scheduling.md` — acceptance baseline and 5-PR roadmap for the Discuss long-session OOM (a production thread composed a ~1.74M-estimated-token context, fully materialized it, and drove the Server past 8 GiB RSS into an OOM kill). Written in the style of `session-runtime-requirements.md`: CM-* requirement IDs, MUST-as-merge-requirement, black-box acceptance scenarios, and a definition-of-done invariant (peak memory ∝ one selected in-budget context × bounded concurrency, independent of total history size).
2. **PR 1 of that roadmap** — the pre-materialization hard boundary and process memory backstop (the P0 stop-the-bleeding layer).

## PR 1 implementation

- **Estimation authority** — `contextfrag.EstimateBytesPerToken` remains the estimation authority (aligned with the merged #1072 envelope work); the new dependency-free `internal/tokenest` anchors its numeric value for the `chat/timeline` and channel estimators, because the `internal/arch` guards forbid those packages from importing `agent/context` (CM-EST-001). The timeline (`chars/2`), application (`len/4`), and fragment estimators all converge on the shared constant — both sides swap together when a real tokenizer lands; no package keeps a private bytes-per-token value.
- **Budgeted composition** — `timeline.ComposeContextWithArtifactsBudgeted` makes the admission decision on merge-entry metadata (lengths of already-held references) *before* materializing anything. Over budget it keeps artifact summaries plus the newest contiguous window, never opens the window on an orphaned tool response, and fails closed with `ProtectedOverflow` when the protected set alone exceeds the budget (CM-ADM-001/002).
- **Byte-budgeted history loads + SQL-side admission measures** — `MeasureActiveMessagesBySession` (SQL-side `COUNT(*)` + `SUM(octet_length(content::text))`) sizes a session without shipping payloads, and the `…WithinBytes` query variants admit rows newest-first on the database side until the content byte budget is spent (the crossing row is kept, so the newest row always loads). Discuss TR loads and legacy history loads run under this bound, so process memory no longer depends on total history size (CM-ADM-001).
- **Server-wide absolute cap** — `[agent] context_absolute_max_tokens` (default 200k, can be raised, never disabled). The effective budget everywhere is `min(model window, cap)`, and the cap alone when the model has no configured window — so `context_window = 0` no longer disables trimming, the provider budget preflight, or compaction (fixes the `providerContextBudgetPlan` nil-plan semantics at its application-level source).
- **Discuss admission, both sides** — the driver composes under the cap with admission logging; the agent side re-admits the composed messages against `min(window, cap)` before SDK conversion and before the ACP prompt concatenation, failing closed with the stable `context.protected_overflow` error code.
- **Degradations shrink, never grow** — artifact-frontier load failures (`loadArtifacts` / `loadTimelineArtifacts`) now degrade into the bounded admission window with a stable `context_admission_degraded` log instead of silently recomposing the full raw history (CM-ADM-003).
- **Process backstop** — compose deployments (production + devenv) set an explicit server memory limit and `GOMEMLIMIT` (~85%), so the Go runtime backs off before the cgroup OOM kill (CM-PRC-001).
- **Observability** — stable structured-log keys `context_admission`, `context_admission_rejected`, `context_admission_degraded` with estimated/selected/budget/dropped fields across the discuss-compose, discuss-turn, discuss-history-load, ACP-prompt, and pipeline paths (CM-OBS-001).

## Verification

Rebased onto `main` after #1072 (`919d1378e`): no conflicts — the `fragment/tokens.go` hunks are disjoint and the alias and envelope functions coexist.

- `go build ./...`
- `go test -count=1` (repo-pinned Go 1.25.6) over both stacks' packages: `internal/tokenest`, `internal/chat/timeline`, `internal/agent/application`, `internal/config`, `internal/channel/discuss`, `internal/arch`, `internal/agent/context/fragment`, `internal/chat/message`, `internal/contextview`, `internal/agent/runtime/native`, `internal/agent/tool`, `internal/models` — all green
- `golangci-lint run` over the touched packages — clean
- Note: running the suite with Go ≥ 1.26 fails the pre-existing `TestRunMidStreamRetrySuppressesRawErrorAfterStepBudgetCancellation` on plain `main` as well (a `context` package behavior change, unrelated to this PR); use the pinned toolchain.


chen-ran:
Revised in 9fdc430 to address six review points:

1. **PR 1 dependency inversion** — admission needed size measures that were scheduled for PR 4. CM-ADM-001 now requires metadata-only measures (SQL-side `COUNT(*)` + `SUM(octet_length(...))` aggregates, never load-to-measure), and that coarse measurement infrastructure is explicitly a PR 1 item; PR 4's persisted per-fragment costs refine it later.
2. **Eviction/replay sequencing** — CM-CCH-001 and PR 3 now carry a hard constraint: bounded (paginated, frontier-first) replay lands before or atomically with cache eviction, never eviction first, to avoid load/evict thrash on cold starts.
3. **Zero-copy immutability** — CM-REP-001 now mandates an enforced read-only contract for shared `json.RawMessage`/`[]byte` payloads: stated convention plus checksum test protection under `-race`; mutation requires copy-first.
4. **Sync-compaction latency** — new CM-CMP-003: backstop fires only at the hard threshold, is measured in turn-latency terms (p50/p95), and rolls out gated with a shadow mode first. PR 2 also notes `runCompactionSync` is not a drop-in for the Discuss corpus — only its single-flight/epoch/cooldown semantics carry over.
5. **Acceptance target** — the arbitrary "< 1 GiB" absolute figure is replaced by the invariant-based assertion (peak RSS at 2× history scale within a small constant factor of 1×) plus a relative reduction vs the recorded pre-fix baseline; absolute bytes are reported, not gated on.
6. **Incremental render correctness** — CM-RPL-002 now requires per-node dirty granularity (edits/deletes invalidate affected nodes and coalesced neighbors) and byte-equality verification against a full re-render oracle via property/fuzz tests.

chen-ran:
`df3132a` closes the P0 review finding: **admission previously ran only after unbounded loads** — the doc marked SQL-side measures as MUST while the checklist deferred them, and the Discuss chain still loaded and decoded every history response since the Unix epoch before any guard.

What changed:

- **SQL-side admission measure** — `MeasureActiveMessagesBySession` aggregates `COUNT(*)` + `SUM(octet_length(content::text))` in the database, so admission sizes a session's history without shipping any payload into the process.
- **Byte-budgeted loads** — `ListActiveMessagesSince[BySession]WithinBytes` admit rows newest-first on the database side via a running window sum until the budget's byte equivalent is spent (the crossing row is kept, so the newest row always loads), returned ascending. (sqlc v1.31.1 quirk: the derived-table alias only resolves when the parameter comparison carries an explicit `::BIGINT` cast.)
- **Every DB-backed history load feeding composition is now bounded by `min(context budget, absolute cap) × bytes-per-token`:** the discuss TR load (was since-epoch unbounded), the pipeline-chat TR load, and legacy `loadHistoryRecords` (session- and bot-scoped). Total history size no longer bounds process memory; the budget does. GOMEMLIMIT is back to being only the last-resort backstop, not a substitute for the boundary.
- **Observability** — discuss history loads log the database aggregate next to what was actually loaded (`context_admission path=discuss_history_load`) whenever the bound cut anything.
- **Doc consistency** — the contradicting checklist item is now checked with the landed description; the remaining unbounded resident state (RC pipeline cache residency, event replay) is explicitly scoped to CM-CCH-001/CM-RPL-001 in PR 3, not claimed by PR 1.

The stacked #1078 has been merged up and re-verified (`go build ./...`, full `go test ./...` via pre-commit, package tests on message/discuss/application/config).

chen-ran:
`436ac9f` aligns this PR with the context-budget landing stack's estimator scheme (#1012, in-flight #1072) instead of forking it:

- **`contextfrag/tokens.go` keeps its original function bodies and remains the named single swap point.** The only remaining change is that its constant aliases `tokenest.BytesPerToken`, and that line sits outside every hunk of #1072. A local test-merge of #1072's head onto this branch now auto-merges cleanly — `ProviderEnvelopeTokens` and the constant anchoring coexist without conflict.
- **`internal/tokenest` is demoted from "authority" to dependency-free anchor.** It exists only because the architecture guards (`TestTimelineOnlyDependsOnTurnPort`, `TestChannelAgentDependenciesStayOnPorts`) forbid chat/timeline and channel packages from importing `agent/context`, while CM-EST-001 requires their estimators to agree with the fragment ledger. Its package doc now says exactly that and credits the origin.
- **The design doc defers overlapping work to the stack**: the PR 4 step-reselection accounting item now defers to #1072 (`ProviderEnvelopeTokens`, hash-only `ProviderPayloadHash`) rather than re-implementing it, and §3.2's representation table marks that row as fixed in flight.

Merge-order note: #1072 landing first is preferred; this branch then rebases with no expected conflict in `tokens.go` (verified by test-merge; `service.go` also auto-merges).

chen-ran:
**`internal/tokenest` is removed** (9ead44e83). Where things landed and why:

- `contextfrag` remains the estimation authority (`EstimateBytesPerToken`, `TokensFromBytes`) and now also owns `DefaultAbsoluteCapTokens`. `contextfrag/tokens.go` is back to byte-identical with `main`.
- The **turn port** re-exports the vocabulary (`turn.ContextBytesPerToken`, `turn.EstimateTokensFromBytes`, `turn.DefaultContextCapTokens`) for `chat/timeline` and `channel`, which the arch guards keep off `agent/context`. `config` → `turn` would cycle (`turn` → `decision/input` → `db` → `config`), so `config` and `application` reference `contextfrag` directly.
- The two structurally identical admission algorithms (`timeline.ComposeContextWithArtifactsBudgeted`, `application.admitDiscussMessages`) now delegate to a single **`turn.AdmitContextEntries`**.

The shared core also fixes a review-confirmed defect: the leading-orphan tool trim could deselect the **protected newest entry**, silently yielding an empty or summaries-only window — a silently skipped turn, a summaries-only run that consumed the cursor, or an invalid provider payload. It now fails closed with `ProtectedOverflow`, and the orphan trim also applies under budget so a byte-budgeted history load that cuts a tool pair in half cannot open the window on the orphaned response. Covered by unit tests at the core and both call sites.

Forwarded into #1078 (clean merge, no `tokenest` references remain there either).

sheepbox8646 (approved):

PR #1078, opened 2026-08-24
feat(context): add sync compaction and bounded timeline replay

chen-ran:
## What this is

This PR now combines PR 2 and PR 3 of `docs/design/context-memory-scheduling.md`:

- the pre-turn synchronous compaction backstop (CM-CMP-001/002/003); and
- bounded timeline replay, incremental projection, and bounded in-process timeline caching (CM-RPL-001/002, CM-CCH-001, CM-OBS-001).

PR #1077 is already on `main`, so its admission/trim boundary remains the hard provider-call safety boundary. Compaction is an optimization before that boundary; failure still falls back to the already-bounded admitted context.

Combining PR 3 here removes the main limitation in the previous description: the synchronous compaction path and cold timeline replay no longer load all raw history into the Go process before applying a budget.

## Synchronous compaction (roadmap PR 2)

**Discuss (native and ACP).** Before a model call, `Service.maybeSyncCompactDiscuss` measures raw compactable pressure (artifact summaries excluded) against 75% of the effective context budget. In active mode it reuses the existing per-session single-flight, epoch, and cooldown protocol. When a summary lands, the current run ends with `DiscussEventRecompose`; the driver reloads artifacts, recomposes, re-admits, and resubmits. Recompose attempts are capped at three and do not advance the discuss cursor.

**Pipeline chat.** The pipeline path reports raw compactable pressure before admission trimming, runs the same gated synchronous backstop, and recomposes in process after a successful summary. The same pressure also feeds the existing asynchronous trigger.

**Failure and manual recovery.** A failed/disabled/cooldown compaction or missing summarizer never permits an unbounded provider call. `/compact run` remains the manual, session-scoped recovery path.

**Rollout gate.** `[agent] sync_compaction = off | shadow | active`, default `shadow`. Shadow emits `sync_compaction_backstop` would-have-fired observations without adding summarizer latency or cost. Legacy history chat keeps its existing synchronous backstop.

## Bounded replay and cache scheduling (roadmap PR 3)

- Timeline replay loads the active compaction frontier first, then reads uncovered events newest-first through a stable `(received_at_ms, created_at, id)` keyset. Covered message/edit payloads are filtered in PostgreSQL before `event_data` crosses into the process. A covered message with a post-frontier edit keeps its full event chain so the edit still has a base node.
- Replay has a 64 MiB process-side byte ceiling and 256-row keyset pages. The pipeline logs admitted replay event count/bytes and persisted event count.
- Runtime compaction uses an oldest-first database byte-admission query before Go selection. Its budget derives from the summarizer input allowance, with a 512 KiB floor and 8 MiB ceiling. If the leading candidate alone exceeds the budget, compaction fails closed as a logged noop instead of loading that payload.
- `ListHistoryTurnsByBot`, which currently has no runtime caller, now has a defensive 512-turn limit. The old unbounded compaction query remains only as a compatibility/diagnostic surface; runtime compaction uses the bounded variant.
- `Pipeline` now reduces owned IC state in place, renders only dirty/new nodes, and detaches rendered slices before edits/deletes so snapshots already queued to discuss workers stay immutable. Public pure `Reduce` plus full `Render` remains the equivalence oracle.
- The in-process timeline cache is bounded by LRU, a six-hour TTL, 256 resident sessions, and approximately 256 MiB of serialized IC+RC bytes. Session deletion/session history reset calls `DropSession`; bot-wide history reset conservatively clears all resident projections.
- Replay/cache counters and stable structured eviction/admission logs expose event count, event bytes, resident bytes, eviction count, and evicted bytes.

The incremental path has deterministic unit/property/fuzz coverage across append, duplicate delivery, edit, delete, rename, and pinned-message sequences, including snapshot immutability and LRU/TTL/resident-byte eviction.

## Compatibility and landing order

- No schema migration or wire/protobuf change is required; only new SQL queries and generated sqlc methods are added.
- The existing split-mode upgrade constraint still applies: until #1120 lands, upgrade Channel before or together with Server. A pre-recompose Channel cannot interpret the content-light recompose signal. Embedded all-in-one deployments are unaffected. #1120 touches adjacent discuss-worker lines, so whichever lands second may need a small rebase.
- `shadow` remains the default even with bounded reads because active synchronous summarization still changes latency/cost and because mixed-version split deployments need the ordering above.

## Verification

- `mise run sqlc-generate`
- `go test ./...`
- `go test -race ./internal/chat/timeline ./internal/agent/context/compaction ./internal/channel/inbound`
- `mise run lint` (passes; UI contract reports the repository's three existing warnings)
- `git diff --check`
- FX validation for both embedded and split server shapes runs in `cmd/agent` tests.

The PostgreSQL integration suite now exercises bounded compaction admission and frontier-aware replay when `TEST_POSTGRES_DSN` is set. This machine had neither a configured DSN nor a running Docker daemon, so those real-PostgreSQL cases were skipped locally; sqlc generation, query compilation, all non-DSN tests, race tests, and lint passed. CI/human QA remains required for the real database path and end-to-end latency behavior.


ChrAlpha:
Pushed a doc commit (64bf93f73) recording the split-mode upgrade ordering constraint: a pre-recompose Channel binary decodes the nil-payload recompose event as a failed stream event, so Channel must be upgraded before (or together with) Server; embedded all-in-one deployments are unaffected. Related hardening landed separately as #1120 — with that guard, a pre-recompose Channel merely fails closed and retries after the summary lands instead of silently consuming the trigger; note #1120 and this PR touch the same worker lines, so whichever lands second needs a trivial rebase.

HoneyBBQ (approved):

ChrAlpha (approved):

PR #1104, opened 2026-08-29
fix(display): share one WebRTC UDP port

### LINKED ISSUE #946: [Bug]: Default WebRTC UDP range spawns 202 docker-proxy processes and consumes ~1 GB RAM
### Bug Description

The default Docker Compose configuration publishes the entire WebRTC UDP port range from 30000 to 30100:

```yaml
MEMOH_DISPLAY_WEBRTC_UDP_PORT_MIN: "${MEMOH_DISPLAY_WEBRTC_UDP_PORT_MIN:-30000}"
MEMOH_DISPLAY_WEBRTC_UDP_PORT_MAX: "${MEMOH_DISPLAY_WEBRTC_UDP_PORT_MAX:-30100}"

ports:
  - "8080:8080"
  - "1455:8080"
  - "${MEMOH_DISPLAY_WEBRTC_UDP_PORT_MIN:-30000}-${MEMOH_DISPLAY_WEBRTC_UDP_PORT_MAX:-30100}:${MEMOH_DISPLAY_WEBRTC_UDP_PORT_MIN:-30000}-${MEMOH_DISPLAY_WEBRTC_UDP_PORT_MAX:-30100}/udp"
 ```
On a standard Linux Docker Engine installation with the default userland-proxy setting enabled, Docker starts one docker-proxy process for every published port and address family.
The default range therefore creates:
101 UDP proxies for IPv4
101 UDP proxies for IPv6
202 UDP proxy processes in total
209 docker-proxy processes in total after including the TCP mappings
1,520 proxy threads
On my 2 GB Debian VPS, these proxy processes consumed approximately 0.8–1.0 GB of physical memory. The complete docker.service cgroup consumed approximately 1.2 GB shortly after boot.
With no swap initially configured, only about 130–150 MB of memory remained available after the services started. A Docker build or another moderate memory spike could then make the entire server, including SSH and the web interface, unresponsive.
No Memoh workload or active WebRTC session is required to reproduce the high baseline memory usage. The proxy processes are created immediately when the default Compose stack starts.
Expected behavior:
The default Docker deployment should not consume approximately half of a 2 GB system's RAM solely for port-forwarding helper processes. The deployment should avoid creating hundreds of userland proxy processes, reduce the default port range, bind only the required address families, or document/configure userland-proxy=false for Linux deployments.

### Steps to Reproduce

1. Use a Linux host with the default Docker daemon configuration. In particular, do not configure `"userland-proxy": false`.

2. Install Memoh from the repository and start the default Docker Compose stack:

```bash
git clone https://github.com/memohai/Memoh.git
cd Memoh
docker compose up -d
```
3.Confirm that the default WebRTC UDP range is published:
```bash
docker port memoh-server | grep udp | head
docker port memoh-server | grep udp | tail
docker port memoh-server | wc -l
```
On my host, memoh-server had 206 published mappings: 202 UDP mappings plus four TCP mappings.

### Error Messages / Logs

```shell

```

### Version

0.17.0

### Model Used

_No response_

### Channel

None

### Additional Context

_No response_

### Contributing

- [ ] I would like to submit a pull request to fix this bug

ZenAlexa:
## What changed

- make the process-level display service own one shared Pion `ice.UDPMux`, and inject that service into both the display HTTP path and the browser tool path
- bind the configured UDP port during server startup and close it through the Fx lifecycle
- change the production and development Compose defaults from `30000-30100/udp` to one configurable `30000/udp` mapping
- retain `MEMOH_DISPLAY_WEBRTC_UDP_PORT_MIN/MAX` as a server-side compatibility fallback; existing Compose overrides migrate their range minimum to the shared port; the new single-port variable takes precedence
- reject direct server configurations that combine the new single-port setting with the legacy range because their runtime priority would be ambiguous
- cover configuration parsing and two simultaneous peer connections gathering host candidates on the same UDP port

## Why this shape

The current Compose defaults expand to 101 UDP mappings. With Docker's userland proxy enabled, that creates the process and memory overhead reported in #946 before a display session even starts.

This follows the ownership model used by Pion's [`ice-single-port`](https://github.com/pion/webrtc/blob/v4.2.12/examples/ice-single-port/main.go) example, [Neko](https://github.com/m1k1o/neko/blob/b0f01cedea68/server/internal/webrtc/manager.go), and [go2rtc](https://github.com/AlexxIT/go2rtc/blob/c245815e75e2/pkg/webrtc/api.go): one long-lived UDP mux is shared across peer connections. Keeping the mux on Memoh's process-level display service also prevents the HTTP handler and browser tool from silently owning separate listeners.

Closes #946.

## Validation

- `mise run lint`
- `go test -race -count=1 ./internal/display`
- `go test -race -count=20 -run '^TestWebRTCAPIReusesSingleUDPPortAcrossPeerConnections$' ./internal/display`
- `go test -count=1 ./internal/display ./internal/handlers ./internal/agent/tool ./cmd/internal/core`
- `internal/container/provider` waits indefinitely for the unavailable local Docker daemon in `TestProvideServiceDockerSlot`; all remaining `./internal/...` packages pass with `-count=1`
- production and development `docker compose config --format json` each render one UDP binding; the previous default rendered 101
- Compose matrix covers the 30000 default, legacy range-minimum migration (32000/32001), and new-variable precedence (33000/33001); each case keeps the host port, container target, and server environment identical

> ⚠️ **No human QA** — this PR has not been verified by a human yet. Remove this line once a human confirms the happy path.


ZenAlexa:
`bc3b186` makes both Compose files migrate the legacy range minimum into the new single-port setting. Existing `MIN/MAX` overrides keep the published and firewalled port and render one mapping. I verified the default, legacy-only, and new-variable-precedence cases for production and development with `docker compose config`.

ZenAlexa:
@codex review

chen-ran (approved):

PR #1085, opened 2026-08-26
feat(context): cross-stack lifecycle, budget, fallback, and ACP context integration

ChrAlpha:
## What changes

Run lifecycles gain their full outcome vocabulary: runtime budget failures and context-view fallbacks now classify live runs into the `failed_budget` and `fallback` statuses the #969 schema forward-declared, with rich outcome classification and enriched, content-light snapshots (model/client, steps, cache usage, kind and trust breakdowns) persisted behind the same terminal fence. Omitted step audits stay nil through JSON round trips, and the lifecycle exit audit survives runtime teardown.

ACP runs stop being lifecycle-blind: prompt context compiles through the contextview pipeline into typed sections with a real manifest and memory-recall trace on the lifecycle holder, and the per-run context budget and tool-exchange policy propagate to tools through the canonical `toolcontext.Session`/`Merge`. The schedule trigger path (#997) uses the same compilation.

Memory recall gains durable traces: recall metadata is retained per run and persisted, and invalidated memory graph builds are fenced instead of racing rebuilds.

The session lifecycle API aggregates run diagnostics (`GET` aggregation over the run-keyed table) with regenerated OpenAPI/SDK contracts.

## Scope

This PR contains lifecycle outcome classification and snapshot enrichment, ACP context compilation and budget propagation, memory recall traces and graph-build fencing, the aggregation endpoint with regenerated contracts, and the deferred-lifecycle acceptance suite.

The tool-roster attribution seam landed earlier via #1041's tool-definition accounting and #1015's run-bound tool context; no web consumer ships in this PR.

This is the final part of the landing work motivated by #890, on top of #959, #968, #1012 (via #1039), #1041, #960, #969, and #1058. It closes out the stack's integration seams.

## Restack notes

The original 99-commit integration branch embedded copies of the PR4 and PR6 chains plus upstream commits from earlier restacks; it was decomposed by matching against every landed stack head and rebuilt as the 16 genuine commits here, standalone on upstream main `fec18e859`. Notable adaptations:

- #1015 upstreamed the shared tool-context infrastructure this branch had been carrying — only the two new fields (`ContextBudgetMaxTokens`, `ContextToolExchangePolicy`) are added to the canonical `toolcontext.Session` and `Merge`, with a merge regression test; the duplicated store/merge helpers were dropped.
- #1046 removed the heartbeat feature — the deferred-lifecycle acceptance tests that pinned `failed_budget`/`fallback` activation moved from the heartbeat vehicle to the schedule vehicle.
[...]

PR #1118, opened 2026-08-31
feat(web): categorized context usage display, budget surfacing, and lifecycle inspector [stack 1/2]

ChrAlpha:
## Summary

Bottom of a two-PR stack. This PR surfaces the fragment-first context orchestration data (#1085/#1099) in the web UI and makes the lifecycle store it reads from safe for long-horizon sessions; #1157 (Trajectory pane) stacks on it, replaces the Phase C inspector dialog with a DeepSeek-Harness-style trajectory, and adds the per-request traces and stored context texts. Review this PR on its own; nothing here depends on #1157.

- **Phase A — categorized Context Usage panel.** Revives the shelved composer context ring + session-info popover as a Claude-Code-style categorized display: the 17 `ContextfragKind`s + tool-def buckets map into 8 fixed-order categories (`context-categories.ts`), rendered as a segmented bar + legend (`context-usage-breakdown.vue`) with an estimate-basis header (`~total / window (%)`). Provider-reported input stays a separate fact row, and the provider-basis single bar remains the fallback for sessions without lifecycle data. The ring shows for native and ACP sessions, stays off for direct runtimes, and refreshes when a turn finishes (one invalidation owner, set-diffed across panes, plus one follow-up refetch once the lifecycle row has landed). It is keyboard reachable: Enter opens the popover with focus inside, the label speaks the usage percentage, Escape returns focus to the ring.
- **Phase B — coherent denominator and compaction mark.** `GET /bots/:id/sessions/:sid/status` returns the persisted `budget_plan` (window, output reserve, tool-def cost, estimator) and `compaction` (`enabled`, `auto_tokens`). The mark derives from the persisted plan window only, only for native runtimes, and only when a summarizer resolves through the trigger's candidate chain; the plan is omitted when the pane's `model_id` resolves to a different provider model than the newest snapshot. The UI uses `budget_plan.window` as denominator, draws the output reserve band and a single auto-compact threshold tick against the window (the trigger's measured quantity differs by turn path, so the tick states the threshold and makes no claim about which segment reaches it). Compact Now and the `/compact` slash command follow the same availability.
- **Phase C — Context Lifecycle Inspector.** A dialog over `/context-lifecycle`: newest-first collapsible turns (first expanded, the reader's expansions survive refreshes), status tone, composition bar, selection counts incl. trimmed, drop reasons with token cost, trust breakdown, window/stable-prefix/cache metrics, mutations, per-step drops, a prompt-diff tag per turn (first turn / system / tools / system+tools / history only), page-scoped aggregates that render only observed values, and `has_more`/legacy notices with a load-up-to-200 action. The inspector chunk loads on first open and its query unmounts with the dialog.

### Long-horizon storage boundary

`selection_decisions` grows with the whole history on every turn, so a 1000-turn session made the list endpoint return 42 MB and `/status` detoast a ~1 MB JSONB per poll. This PR moves that cost out of every hot path:

- `context_lifecycles.snapshot` is now a bounded summary (about 3 KB per row); the per-fragment audit lives in a new `selection_decisions` column. Every write statement splits the incoming snapshot at the storage boundary (`snapshot - 'selection_decisions'` / `-> 'selection_decisions'`) and uses `COALESCE` so a summary-only rewrite can never erase an existing audit. Writes return identity columns only, and reconciliation read-backs no longer project the audit.
- `SelectionTrace` gains `trimmed` and `drop_reason_tokens`, rolled up when the snapshot is built, so no reader needs the per-fragment audit. Assistant-message metadata copies persist the summary only.
- Migration `0147` adds the column, then splits and back-fills existing rows per team (it opens and restores the `teams` policy like `0120`; it holds an exclusive lock on `teams` for its transaction, so run it with the server stopped as the compose stack does).
- `/status` reads one summary row, the list projects summaries, the per-run detail endpoint is gone, lifecycle queries never persist to localStorage, and a storage quota error pauses cache persistence for the page instead of retrying on every change.

Measured on a 1000-turn seeded session on the dev stack: `/status` 20 ms → 7 ms; `/context-lifecycle?limit=50` 42 MB / 463 ms → 297 KB / 12 ms (Postgres 1,271 shared buffers → 210); inspector open 0.7 to 4.5 s with long tasks → about 100 ms with none.

### Trade-offs kept

- The audit column still grows with history on disk; retention (TTL or last-K runs per session) is a separate decision and is one `UPDATE` on that column.
- Runs whose run-table row was never written recover from the summary-only metadata copy and therefore carry no per-fragment audit.
- Pages beyond 200 turns need cursor pagination, not included here; `steps`/`mutations` per row are step-bound, not history-bound.
- After a manual Compact Now the display reflects the last completed turn until the next turn produces a snapshot.

### Latest rebase and live validation

- Rebased onto `origin/main` at `d2058705e`. Kept main's composer model-preference and mic/send behavior, preserved the direct-runtime context-ring gate from the original merge history, and regenerated SDK conflicts from the merged OpenAPI definition.
- Rolled the lifecycle split migration to `0147` because main now uses `0146` for session model preferences; regenerated sqlc and passed lifecycle down/up and database-owner migration tests.
- Fixed two issues found in the real development UI: the inspector now restores keyboard focus after the dialog releases its focus trap, and the usage popover correctly constrains its scroll viewport so bottom actions remain reachable.
- Docker Desktop dev stack with the current contract-v3 workspace image: 4 real `mimo-v2.5-pro` turns, a real `pwd` tool call returning `/data`, manual compaction, and successful marker recall after compaction. Checked empty-session fallback, usage/API reconciliation, summary-only lifecycle responses, turn-end refetch plus one follow-up, keyboard flow, desktop dark/light and 390px Chinese layout.
- Full Go tests, focused Web tests (111), lint and Web build pass. Web typecheck has 446 diagnostics identical to a fresh same-dependency main worktree. Full Vitest has 4 failing files / 5 failing tests, all reproduced on main (desktop URL expectations/runtime build, media-gallery storage setup, preview i18n mock).
- This run did not repeat the earlier 1000-turn performance test or live ACP/direct-runtime flows. The inspector remains the self-contained surface in this PR; #1157 folds it into the Trajectory pane.

## Test plan

- Unit (Go): fragment rollup and `Summary()`, handler composition, plan applicability, compaction mark derivation, legacy status fallback, compaction threshold policy.
- Unit (web): category mapping, view adapters incl. prompt diff, breakdown geometry, inspector turns, turn-end invalidation, cache persistence breaker, session context view.
- Postgres integration (`go test -tags integration`): query round-trip with the split, summary-only rewrites keep the audit, `0147` down/up round-trip on a pre-split row incl. the token rollup, migration as a non-superuser database-owner role.
- Live e2e (Playwright against the dev stack with a real provider turn): popover numbers reconcile with `/status` exactly, no `selection_decisions` on the wire, the inspector opens without per-turn detail requests, keyboard path ring → popover → inspector → back, turn end refetches once plus one follow-up.

## QA

1. `mise run dev`, open a session with at least one assistant turn.
2. Hover the ring left of the model pill, or focus it and press Enter: categorized bar + legend, `~` totals, provider input row, output-reserve band and free space when the window is known.
3. Click **Context Inspector**: per-turn composition, selection and drop reasons, trust, metrics, prompt-diff tags; the dialog stays open while the popover closes; Escape returns focus to the ring.
4. Sessions without lifecycle data fall back to the single-bar display and an empty inspector.


ChrAlpha:
## Long-conversation e2e + CI failure analysis

**Widget stability (dev stack, real turns).** Drove one session to 14 lifecycle turns (28 messages) through the local channel API and the composer. The panel and inspector stayed correct and stable throughout: per-turn totals reconcile with `/context-lifecycle` snapshots at every step; the 14-row inspector list renders newest-first with the first row expanded and supports multi-expand; two full dialog open/close cycles plus three popover hover cycles produced zero console errors from the new components (the only console noise is a pre-existing `/api/connectors/catalog` 502 from the dev Connect-It container). Dynamic category appearance verified live: a `Memory` row (146 tokens) joined the legend as memory recall kicked in mid-conversation.

**Upstream finding surfaced by the widget (not addressed in this PR).** On this stack (base `a23d24a1f` + these display-only commits), the native pipeline-chat path composes follow-up turns without real conversation content:

- `conversation_event` stays at 0 tokens / ≤9 text_bytes across all 14 turns even though in-window history rows exist (verified in `context_lifecycles.snapshot->breakdown`);
- a 80,500-byte user message composed as `current_user_message: 56` — the same size as the short messages before it — and the model's reply ("紫") answered the *previous* turn's theme, ignoring the actual request;
- a recall probe ("list the numbers I mentioned in earlier turns") failed behaviorally; replies across turns look driven by `memory_recall` (which grew 103→146) rather than history.

Message-row shapes: user rows persist string `content` (long-standing shape, June sessions identical), assistant rows persist reasoning-only parts for this `openai-responses` model. `messageconv.ModelMessageToSDKMessage` handles string content correctly, so the drop happens deeper in the pipeline history/current-message plumbing. None of the code in this PR touches composition (both review passes pinned the diff to read-side handlers + frontend), and the widget is what made the anomaly measurable — filing separately after bisecting against plain `main`.

**CI: the `ESLint / Test` failure is unrelated to this diff.** `docker/server-entrypoint.test.ts > docker backend…` fails with exit 143: the test spawns the real entrypoint with a 10s `spawnSync` timeout, and the entrypoint's `docker run alpine:3.23` probe pays a cold-registry pull on the runner; the job stderr shows the pull completing and the process being killed at the deadline. The job is path-triggered (`apps/**/*.ts` matches this PR), `docker/` is untouched here (last change: #918, 2026-08-02), and the same suite passes locally in 2.44s with a warm image. Two reruns hit the same slow-pull window. Cheap durable fixes, happy to send either as a separate PR: pre-pull `alpine:3.23` in the workflow step before `pnpm test --run --project docker`, or raise that one spawn timeout.

chen-ran (approved):

PR #1123, opened 2026-08-31
feat(apperror): make context budget errors actionable

ChrAlpha:
## What this is

A turn rejected for context budget surfaced only "Required context exceeds the model context budget." — accurate, but with nothing the user can do about it, in English, on every surface.

## How it works

- The Detail strings for `context.protected_overflow` and `context.budget_unsatisfied` now name the recovery paths: run `/compact` to summarize older history, shorten the request (window-too-small only), or switch to a model with a larger context window. Channel replies use the Detail directly; the web UI localizes by code.
- All three web locales updated per the `codesync(error-catalog)` contract; pinned expectations updated in Go tests and the web vitest suite.
- Reviewed: a repo-wide sweep found no other consumer of the old strings; `/compact` is registered on the shared channel command group and shipped as a web chat quick action; channel error rendering applies secret redaction only, no length caps. For a reserve-driven `budget_unsatisfied` (window smaller than the output reserve) only the model switch helps — the wording hedges all three remedies.

## Verification

- `go test ./internal/apperror/ ./internal/agent/runtime/native/ ./internal/handlers/ -count=1`; `pnpm --filter @memohai/web exec vitest run src/utils/api-error.test.ts` (51/51)
- Full `go test ./...` + lint via pre-commit; adversarial review pass


ChrAlpha:
Folded into #1122 (commit 7f4b5cda9, cherry-picked verbatim): both changes serve the same context-budget failure path — adaptive retry when compaction fails, actionable copy when the turn is ultimately rejected. All targeted tests re-run green on the combined head.

PR #1266, opened 2026-09-14
fix(chat): 把 (turn_id, role) 唯一这条不变量钉住，破坏时不再静默丢 turn

chen-ran:
## Author

- [ ] Human
- [x] Agent

## Type

- [x] bug
- [ ] feat
- [ ] test

## Summary

The web transcript pairs a live or optimistic turn with its settled twin by `(turn_id, role)`, one to one: `adoptRenderIdentity` hands each on-screen render key to exactly one incoming turn. Two problems, both raised in review on the previous revision:

**1. Positional pairing still lost messages and minted duplicate keys.** Consuming each prior once with `shift()` does not guarantee the resulting render ids are unique. Reproduced: the screen holds `a3`, history returns `a1` and `a3` — `a1` inherits `a3`'s id, the trailing `a3` keeps its own id `a3`, and both end up under the same key. The id-keyed merge collapses them into one turn, `replaceMessages` emits a duplicate Vue key, and a later refresh that returns only `a3` overwrites `a1`'s content.

Matching now runs strongest-first: a row the database has named is the same row wherever it appears, so its stored id decides before anything else; only the leftovers pair by `(turn_id, role)`, and only against turns with no stored id (the live and optimistic ones whose key a settled twin is meant to inherit). A pairing that would collide is skipped and the twin keeps its own id.

**2. The invariant was not actually guarded.** The previous Go test hand-built five row sets that already satisfied the constraint and fed them to the converter. It never went through persistence, SQL, or the paging API, so a persistence path that started writing a duplicate user identity would leave those fixtures passing.

`TestPostgresHistoryPageKeepsTurnRoleIdentityUnique` now drives the real chain: persist through the message service (admission-style supplied turn identity, a mid-run user input, a following round) → read back through the paged UI history query → reproduce the handler's `reverseMessages` → convert. It asserts `(turn_id, role)` uniqueness, that a mid-run input opens its own turn numbered after the request, and that the page is ordered by turn position. Page sizes 2 / 4 / 8 / 30 are covered. `migrations.yml` runs `./internal/agent/view` in the DSN step.

The original converter test is kept, with its claim corrected: it guards the converter's own grouping rules (change `flushPending`'s triggers and the same rows can produce two turns under one id), not persistence.

This is defensive hardening, not a regression fix. No current persistence path is known to produce multi-segment turns — `persistHistoryTurn` opens a new turn for every visible user row, tail linking is restricted to assistant and tool rows, and the one query that could file a message under an existing turn has no Go caller.

## Related Issues

Follows #1264 and #1265.

## Validation

Automated, all run locally:

- `internal/agent/view` unit + integration tests against a live PostgreSQL 18 container (`migrate up`, then `TEST_POSTGRES_DSN=…`). Verified the new guard actually bites: temporarily filing a second visible user row under an existing turn made it fail with `turn 2 (turn-1/user) repeats the identity already used by …`; the probe was removed afterwards.
- `apps/web` vitest: 1773 passed. Verified the three new pairing tests are not vacuous — reverting `adoptRenderIdentity` to the previous implementation made two of them fail with `expected [...] to have a length of 2 but got 1` (lost turn) and `expected 1 to be 2` (duplicate key).
- `go build ./...`, `go vet`, `gofmt`, `eslint`, `vue-tsc`.

**Not run: real UI verification.** See below.

## Screenshots / Recordings

The development environment could not be brought to a usable state, so no screenshot of the affected flow exists.

What was attempted, in order:

1. `mise run dev` — the workspace image build runs for >10 minutes (full desktop environment); the compose services came up alongside it.
2. `memoh-dev-web` was crash-looping on `check-patched-deps.mjs` (`reka-ui@2.10.1`: "directory is named as patched but holds unpatched content"). Applied the recovery the error message prescribes, scoped to the dev container volume, then reinstalled: `✓ patched dependencies verified against installed content`. Vite came up on :18082 and the login page rendered.
3. `memoh-dev-server` exits at startup with `column "bot_agent_id" does not exist (SQLSTATE 42703)` — the dev database schema is behind the code. Restarting the container recompiles from the mounted source via `air`, and the failure persists.
4. Running migrations against the dev database fails: `could not read blocks 0..0 in file "base/16384/21498": read only 0 of 8192 bytes`. The `schema_migrations` relation's heap file is physically corrupted. Every other table reads normally (`bots` 2, `users` 1, `models` 4, `providers` 1, `bot_sessions` 4, `bot_history_messages` 14).

Repairing this means recreating `schema_migrations` at a version that has to be inferred, or resetting the dev Postgres volume — which would destroy the environment's bots, sessions, history, and the configured provider credential. That is the environment owner's decision, not one to take while verifying a PR, so verification stops here and this PR should not merge on automated evidence alone.

## Human QA

- [ ] Human QA passed


chen-ran:
## 回应 review：两条都确认属实，已修（`91946af`）

### 1. 按位置配对仍会丢消息、产生重复 key —— 已修

复现确认。`shift()` 只保证旧对象被认领一次，不保证最终渲染 ID 唯一：屏幕上是 `a3`，历史返回 `a1`、`a3` —— `a1` 继承了 `a3` 的 ID，后一个 `a3` 保留自己的 ID `a3`，两个渲染 ID 都成了 `a3`。折叠和重复 key 照旧。

改为**按强度从高到低匹配**，正是 review 建议的顺序：

1. 数据库已经命名过的行，无论出现在哪里都是同一行，先用它的 stored id 决定；
2. 剩下的才按 `(turn_id, role)` 配对，且**只与尚无 stored id 的那些 turn**（live / optimistic，本来就是要把 key 交出去的一方）配；
3. 最后一道闸：不可重复发牌——会撞车的配对直接跳过，让 twin 保留自己的 ID。

review 提的第三种情形（「下一次只刷新 a3 会把 a1 内容覆盖成 a3」）也随之消失：`a3` 现在按 stored id 精确命中自己。

新增 3 条用例，并验证过非空洞——把实现临时换回 `shift()` 版本后，其中 2 条分别以 `expected [...] to have a length of 2 but got 1`（丢 turn）和 `expected 1 to be 2`（重复 key）失败。

### 2. Go 测试没有真正守住持久化不变量 —— 已修

这条批评完全成立，我原来的说辞是过头的。固定样例确实只能守转换器自己的分组规则。

新增 `internal/agent/view/turn_identity_postgres_integration_test.go`，跑**真实链路**：

- 用 message 服务落库（准入式的指定 turn 身份 + mid-run 用户输入 + 后续轮次）
- 经**分页 UI 查询** `ListLatestUIBySession` 读回
- 复刻 handler 的 `reverseMessages`
- 再交给转换器

断言：`(turn_id, role)` 唯一、steer 自开 turn 且编号在请求之后、整页按 position 升序。分页覆盖 `limit` 2 / 4 / 8 / 30。

**验证过它会咬**：临时往已有 turn 里塞第二条可见用户行，测试按预期失败（`turn 2 (turn-1/user) repeats the identity already used by ...`），验证后移除。

`migrations.yml` 已把 `./internal/agent/view` 纳入带 DSN 的那一步，并把 `internal/agent/view/*postgres_integration_test.go` 加进 path filter。本地是起 `postgres:18-alpine` + `migrate up` 跑通的。

原来的转换器测试保留，但改掉了它的说辞：它守的是转换器自己的分组规则——改 `flushPending` 的触发条件、改什么开系统 turn，同样的行就可能变成两个 turn——**不声称**能守持久化。

### 关于「不能把它说成已发生的生产回归」

同意，PR 描述里我原本也写了「尚未证明当前正常持久化路径会产生这种多段数据」。现在有了真实链路的守卫，这句话的定位更清楚了：**这是防御性加固，不是回归修复**。

